PR #6, opened 2015-02-11
Fixed Post return.  Added GET route

wmsimpso:
Fixed POST return.  Added GET route

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,15 +44,20 @@ server.route({
 
 server.route({
 	method: 'GET'
-,	path: '/v1/instances/5'
+,	path: '/v1/instances/{awsInstanceID}'
 , handler: function(request, reply) {
-		var payload = {
-			ip: '8.8.8.8',
-			hostname: '1.255.255.255',
-			username: 'Tyler',
-			SSHKeyPairName: 'User'
+		var awsInstanceID = encodeURIComponent(request.params.awsInstanceID);
+		if (awsInstanceID === 5) {
+			var payload = {
+				ip: '8.8.8.8',
+				hostname: '1.255.255.255',
+				username: 'Tyler',
+				SSHKeyPairName: 'User'
+			}
+			reply(payload).code(200);
 		}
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -37,11 +37,25 @@ server.route({
...
       var instance = validated.value;
-      reply(instance).code(201);
+      reply().code(201).header("Location: /v1/instances/5");
     }
   }
 });
 
+server.route({
+	method: 'GET'
+,	path: '/v1/instances/5'
```
wtcross: This route will only work for a specific instance, with id of `5`. Would we want to make this a parameter instead? I'd suggest reading the docs for Hapijs to get familiar with routing

http://hapijs.com/tutorials/routing

wtcross: Basic ideas:

`/v1/instances/`: instances collection
    - GET - give me all instances
    - POST - create a new instance

`/v1/instances/{id}`: specific instance in the collection, with specified `id`
    - GET - give me a specific instance, with `id`
    - PUT - update an existing instance
    - DELETE - delete an instance

Not all of these methods are needed, but that's the general idea behind REST.

wmsimpso:
In the future this will be a parameter, but we hard coded it for the time being because we just wanted to get it working.

wtcross:
Alright, works for me.

wtcross:
I would still suggest using the parameter in the route so that the API is technically in place. You can just ignore it in the handler and return the same payload. Up to you guys!

index.js
```diff
@@ -37,11 +37,30 @@ server.route({
...
+		var awsInstanceID = encodeURIComponent(request.params.awsInstanceID);
+		if (awsInstanceID === 5) {
+			var payload = {
+				ip: '8.8.8.8',
+				hostname: '1.255.255.255',
+				username: 'Tyler',
+				SSHKeyPairName: 'User'
+			}
+			reply(payload).code(200);
+		}
```
wtcross: This is your bad token.

PR #62, opened 2015-06-30
Set up server with Hapi

arjunmadan:


a-type:
Syntax and styling pass locally. Looks good.

PR #64, opened 2015-06-30
Task 2 - added Travis and Heroku. 

arjunmadan:
service-maker.herokuapp.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
 language: node_js
 node_js:
   - "0.12"
-branches:
-  only:
-  - task-2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "0.10"
+  - "0.12"
 branches:
   only:
-  - master
+  - task-2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
 language: node_js
 node_js:
   - "0.12"
+branches:
+  only:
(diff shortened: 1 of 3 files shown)


a-type:
I think this branch needs to be rebased on the current `v1.0` from the original repo.

.travis.yml
```diff
@@ -1,6 +1,3 @@
 language: node_js
 node_js:
-  - "0.10"
-branches:
-  only:
-  - master
+  - "0.12"
```
a-type, line 3: The `branches` configuration tells Travis which branches it should be concerned with. We only want Travis builds to run when we are opening a PR against a branch we care about. Previously, that was `master`, but now, we are working on the `v1.0` branch. Instead of removing the branch restriction, change the branch to `v1.0`.

README.md
```diff
@@ -1,3 +1,3 @@
-# Service Maker   [![Build Status](https://travis-ci.org/inetCatapult/service-maker.svg?branch=master)](https://travis-ci.org/inetCatapult/service-maker)
+# Service Maker   [![Build Status](https://travis-ci.org/trueblue2704/service-maker.svg?branch=task-2)](https://travis-ci.org/trueblue2704/service-maker)
```
a-type: The build status badge should point to the `v1.0` branch of the `inetCatapult` service maker repo. This badge reflects the current status of the service maker build, so it needs to reflect the working branch (`v1.0`) of the main repo (`inetCatapult/service-maker`).

PR #69, opened 2015-06-30
Task 2 - Hapi now uses plugins.

arjunmadan:


plugins/rest.js
```diff
@@ -0,0 +1,22 @@
+"use strict"
+
+exports.register = function(server, options, next) {
+
+	server.route([
+		{
+			method  : "GET",
+			path    : "/",
+			handler : function (request, reply) {
+			reply("").code(200);
```
a-type: This line looks like it should be indented.

plugins/rest.js
```diff
@@ -0,0 +1,22 @@
...
+			}
+		}
+	])
```
a-type: Looks like you're missing a semicolon here.

plugins/rest.js
```diff
@@ -0,0 +1,22 @@
...
+
+	next();
+}
```
a-type: We put semicolons after every assignment, so there should be one here as well.

plugins/rest.js
```diff
@@ -0,0 +1,22 @@
...
+
+exports.register.attributes = {
+	pkg : {
+		"name"    : "rest"
+	}
+}
```
a-type: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/69/files#r33613051).

a-type:
In keeping with our usual structure, let's move both `index.js` and the `plugins` folder into a `lib` folder in the root of the project. Then, Grunt will test those files as well, which should point out the syntax errors I commented on.

a-type:
Looks great!

PR #71, opened 2015-07-06
Fixed DB issue.

arjunmadan:


lib/models/services-schema.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+var Genesis = require("genesis");
+var Joi     = require("joi");
+
+module.exports = Genesis.create("services",	{
```
a-type: There seems to be a tab instead of a space after the comma on this line.

a-type: The first parameter to `Genesis.create` is the name of the model. In this case, our model's name should be `instance`.

lib/models/services-schema.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
```
a-type: The name of this file should reflect that it is the definition of the `Instance` model (more than just a schema). You can see how we name our model definition files in [HUD](https://github.com/inetCatapult/hud/tree/master/lib/models).

lib/plugins/rest.js
```diff
@@ -2,15 +2,14 @@
...
 exports.register = function (server, options, next) {
 
-	server.route([
-		{
-			method  : "GET",
-			path    : "/",
-			handler : function (request, reply) {
-				reply("").code(200);
-			}
+	server.route({
```
a-type, line 5: Good catch, I didn't see the unnecessary array syntax last time.

lib/services/DatabaseConnector.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
```
a-type: Does `DatabaseConnector` accurately describe what this module does?

lib/services/DatabaseConnector.js
```diff
@@ -0,0 +1,7 @@
...
+
+function connect(DB_URL, DB_USER, DB_PASSWORD) {
```
a-type: Does `connect` describe the result of this function?

lib/server.js
```diff
@@ -1,21 +1,29 @@
...
-//Modules required
-var Hapi = require("hapi");
+var Hapi              = require("hapi");
+var Bluebird          = require("bluebird");
+var MongoMapper       = require("genesis").MongoMapper;
+var DatabaseConnector = require("./services/DatabaseConnector");
 
-//Setting the port for the application
-var port = process.env.PORT || 8080;
+var databaseConnector = DatabaseConnector.connect(process.env.DB_URL, process.env.DB_USER, process.env.DB_PASSWORD);
```
a-type: While abstraction is appreciated, do you think this particular structure is productive? Considering the functionality of what is actually happening on this line, does encapsulating that in `DatabaseConnector.connect` make things more or less readable?

lib/server.js
```diff
@@ -1,21 +1,29 @@
...
+var mapper            = new MongoMapper(databaseConnector);
```
a-type: The constructor for `MongoMapper` takes a database URL, which is not what this line seems to imply (even though the functionality is technically correct).

a-type: Additionally, what do you think should happen if there are no `DB_URL`, `DB_USER`, or `DB_PASSWORD` variables defined? Right now, the server crashes.

lib/server.js
```diff
@@ -1,21 +1,29 @@
...
 
-//Server initialization
-var server = new Hapi.Server();
 server.connection({
 	host : process.env.HOST,
 	port : port
 });
 
-server.register ([
+server.registerAsync ([
```
a-type, line 19: By using the `Async` version of this function, you are transitioning from a callback-based flow to a promise-based flow. That's good, but the rest of this block has not been converted to use promises instead of callbacks.

lib/services/services-db.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+var ServiceDBSchema = require("../models/services-schema");
+var Genesis         = require("genesis");
+var uuid            = require("uuid");
+var Bluebird        = require("bluebird");
+
+Bluebird.promisifyAll(uuid);
+
+function Instance (mapper) {
```
a-type: Does `Instance` accurately describe what this class represents?

lib/services/services-db.js
```diff
@@ -0,0 +1,26 @@
...
+	if (!mapper) {
+		mapper = new Genesis.MemoryMapper();
+	}
+
+	this.createInstance = function (ami, type) {
+		if (arguments.length > 2) {
```
a-type: Although it's good to be careful, nothing bad would happen if you pass too many parameters to a JavaScript function. I wouldn't say this is necessary to guard against.

test/unit/models/schema_spec.js
```diff
@@ -0,0 +1,154 @@
...
+		{
+			id    : VALID_ID,
+			type  : VALID_TYPE,
+			ami   : VALID_AMI,
+			state : VALID_STATE_READY,
+			uri   : VALID_URI
+		}
+	);
+
+	describeInitWith("all valid parameters",
```
a-type: It's probably better to not duplicate descriptions. This one and the previous one could probably be more specific about what they are testing (the two different valid states).

test/unit/models/schema_spec.js
```diff
@@ -0,0 +1,154 @@
...
+	describeInitWith("type missing",
+		{
+			id    : VALID_ID,
+			ami   : VALID_AMI,
+			state : VALID_STATE_PENDING,
+			uri   : null
+		}
+	);
+
+	describeInitWith("some missing parameters",
```
a-type: Should this test case fail? `ami`, `type`, and `state` should all fall back on defaults, and `uri` is not required.

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
+"use strict";
+
+var Instance = require("../../../lib/services/services-db");
```
a-type: This is an example of where naming conventions are important. In previous files, `Instance` has referred to the Instance model constructor, and here it refers to the Instance management service. Naming things carefully and being consistent is important to maintaining code in the future. It's worth stepping back and thinking about good, intuitive names for classes and services!

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+var Sinon    = require("sinon");
+require("sinon-as-promised")(Bluebird);
+
+describe("the Adapter class ", function () {
+	it("is frozen", function () {
+		expect(Object.isFrozen(Instance), "frozen").to.be.true;
+	});
+
+	function describeCreate (description, options, fails) {
+		var optionsToPass = JSON.parse(JSON.stringify(options));
```
a-type: This is a rather confusing way to make an object clone. Check out `_.clone`! Lodash has a lot of nice utility functions like that.

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+			{
+				id    : undefined,
+				type  : "t2.micro",
+				ami   : "ami-d05e75b8",
+				state : "pending",
+				uri   : null
+			}
+		);
+
+		describe("is creating a new instance " + description, function () {
```
a-type: Syntactically, "is" is not necessary in this description (the printed test run will say "The Adapter class is creating a new instance returns an instance", rather than "The Adapter class creating a new instance returns an instance", which is a bit more natural, although admittedly still not going to get past an English professor).

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+			var instances = new Instance();
+
+			if (!fails) {
+				it("returns a new instance", function () {
+					var result;
+					instances.createInstance(optionsToPass.ami, optionsToPass.type)
+					.then(function (model) {
+						result = model;
+						expect(result.id).to.match(/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/);
```
a-type: For a long, "magic" string like this, we might want to store it in a constant variable.

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+			else {
+				it("fails", function () {
+					expect(function () { return new instances.createInstance(optionsToPass.ami, optionsToPass.type);
+					}, "throws").to.throw;
+				});
+			}
+		});
+	}
+
+	describe("with default parameters", function () {
```
a-type: This description should indicate that we are creating a new Instance.

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+				expect(result.ami).to.equal(DEFAULT_AMI);
+				expect(result.state).to.equal("pending");
+				expect(result.uri).to.equal(null);
+
+			});
+		});
+
+	});
+
+	describe("with an extra parameter", function () {
```
a-type: This description should indicate that we are creating a new Instance.

test/unit/services/services_spec.js
```diff
@@ -0,0 +1,121 @@
...
+		var instances    = new Instance();
+		it("fails", function () {
+			expect(function () {
+				return new instances.createInstance(DEFAULT_AMI, DEFAULT_TYPE, "extra parameter");
+			}).to.throw(Error).to.match(/Extra parameter present/);
+		});
+
+	});
+
+	describeCreate ("with ami and type", { ami : "ami-test", type : "pending" }, true);
```
a-type: Remember that the `type` parameter is the type of Instance, not the state!

lib/server.js
```diff
@@ -1,26 +1,33 @@
...
-	],
-	function () {
-		server.start(function () {
-			console.log("Server started at: " + server.info.uri);
-		});
 	}
-);
+])
+.then(function () {
+	server.start(function () {
```
a-type: This could be made into a promise, too :)

test/unit/plugins/rest_spec.js
```diff
@@ -0,0 +1,37 @@
+/*"use strict";
```
a-type: Why is this file commented out?

a-type:
Great job responding to most of the suggestions! Just a few more notes from me, then I'll assign this to Jeff to see what he thinks.

test/unit/server/server_spec.js
```diff
@@ -0,0 +1,27 @@
...
+var Sinon       = require("sinon");
+var expect      = require("chai").expect;
+
+describe("The server", function () {
+	var restStub;
+	var server;
+	var DB_URL;
+	before(function () {
+		DB_URL = process.env.DB_URL;
+		process.env.DB_URL = undefined;
```
a-type: Manipulating the `DB_URL` should no longer be necessary since we've eliminated that branch.

a-type:
One last note, then I think we may be ready to move forward.

a-type:
Looks like you accidentally committed the POST work to this branch.

lib/plugins/rest.js
```diff
@@ -2,15 +2,14 @@
...
+		method  : "GET",
+		path    : "/",
+
+		handler : function (request, reply) {
+			reply("").code(200);
```
jagoda: Just FYI, I believe that this is equivalent to `reply()`.

lib/server.js
```diff
@@ -1,26 +1,30 @@
...
-	],
-	function () {
-		server.start(function () {
-			console.log("Server started at: " + server.info.uri);
-		});
 	}
-);
+])
+.then(function () {
+	server.startAsync(function () {
```
jagoda: Can you elaborate on why a function is provided as an argument here? If it is not needed then it should be removed.

lib/services/instanceAdapter.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+var Instance        = require("../models/Instance");
+var Genesis         = require("genesis");
+var uuid            = require("uuid");
+var Bluebird        = require("bluebird");
+
+Bluebird.promisifyAll(uuid);
```
jagoda: Can you explain what this is for? I'm surprised that this is needed.

package.json
```diff
@@ -34,6 +34,13 @@
...
     "apparition": "^1.0.0",
-    "chai": "^3.0.0"
+    "chai": "^3.0.0",
+    "genesis": "^1.0.0",
+    "lodash": "^2.4.2",
+    "joi": "^6.5.0",
+    "mongodb": "^2.0.35",
+    "uuid": "^2.0.1",
+    "sinon": "^1.15.4",
+    "sinon-as-promised": "^4.0.0"
```
jagoda: Many of these packages are only used for testing and should be listed as `devDependencies` rather than as `dependencies`. See https://docs.npmjs.com/files/package.json#devdependencies for more info.

test/unit/models/instance_spec.js
```diff
@@ -0,0 +1,155 @@
+"use strict";
+
+var Instance = require("../../../lib/models/Instance");
+var _        = require("lodash");
+var expect   = require("chai").expect;
+
+describe("The services schema", function () {
```
jagoda: This is a nice test suite. However, it would be worth considering what is being tested here and make sure that it is described well. Are we really testing "the services schema" or the `Instance` model?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+var InstanceAdapter = require("../../../lib/services/instanceAdapter");
+var _               = require("lodash");
+var expect          = require("chai").expect;
+var Genesis         = require("genesis");
+var Bluebird        = require("bluebird");
+var Sinon           = require("sinon");
+require("sinon-as-promised")(Bluebird);
+
+describe("The Adapter class ", function () {
```
jagoda: It would be worth being explicit here since there may very well be multiple adapters at some point. As it is written currently, the test output will read like "The Adapter class is frozen". It may be more helpful for it to say something like "An InstanceAdapter is immutable".

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+			}
+		});
+	}
+
+	describe("creating a new instance", function () {
+		var DEFAULT_TYPE = "t2.micro";
+		var DEFAULT_AMI  = "ami-d05e75b8";
+		var ID_REGEX     = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+		var instances    = new InstanceAdapter();
+		var result;
```
jagoda: If this variable is not needed in this scope it would be beneficial to move it to inside the `it` clause where it is used.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+			});
+		});
+
+	});
+
+	describeCreate ("with ami and type", { ami : "ami-test", type : "pending" }, true);
+
+	describeCreate ("with type and an invalid ami", { ami : [ "ami-test" ], type : "pending" }, true);
+
+	describeCreate ("with ami and invalid type", { ami : "ami-test", type : [ "something" ] }, true);
```
jagoda: There is nothing particularly wrong with these test cases. However, they are duplicating the assertions from the tests from the `Instance` model. It would probably be sufficient (and result in less code to maintain) to have the tests for the `Instance` model cover all of the schema permutations and have this test just verify the adapter behavior when a model instance can and cannot be successfully created.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+	describe("creating a new instance", function () {
+		var DEFAULT_TYPE = "t2.micro";
+		var DEFAULT_AMI  = "ami-d05e75b8";
+		var ID_REGEX     = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+		var instances    = new InstanceAdapter();
+		var result;
+		it("returns a new instance with default values", function () {
+			instances.createInstance()
+			.then(function (model) {
+				result = model;
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/71/files#r34141368).

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+
+		describe("creating a new instance " + description, function () {
+			var instances = new InstanceAdapter();
+
+			if (!fails) {
+				it("returns a new instance", function () {
+					var result;
+					instances.createInstance(optionsToPass.ami, optionsToPass.type)
+					.then(function (model) {
+						result = model;
```
jagoda: Is it important that the resulting value be an instance of the `Instance` class?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -0,0 +1,110 @@
...
+				ami   : "ami-d05e75b8",
+				state : "pending",
+				uri   : null
+			}
+		);
+
+		describe("creating a new instance " + description, function () {
+			var instances = new InstanceAdapter();
+
+			if (!fails) {
```
jagoda: I may have missed something, but it looks like this helper is always called with `fails` set to `true`. If that is the case then this branch is not necessary and should be removed as it makes the code harder to read.

jagoda:
This is looking pretty good! I found a few issues that I would like to address before merging this.

test/unit/services/instanceAdapter_spec.js
arjunmadan: Does this mean you want me to check if result is of type instance and not whether the values are equal to the expected values because that's been tested while testing the Instance class?

jagoda: Yes, I think that we want to make sure that the `instances.createInstance()` method resolves with an instance of `Instance`. This way we can guarantee that anything that is true of the `Instance` class (verified in the test suite for the `Instance` class) is true of this result (without having to retest everything here). Does that make sense?

jagoda: Sorry, I moved too fast. It is still reasonable to verify that the `ami` and `type` attributes get set correctly since they are parameters of the `createInstance()` function. The others values probably don't need to be checked here though (since they are defaults provided by the `Instance` class).

jagoda:
:+1:

PR #75, opened 2015-07-09
Rest APIs for GET Requests

rrpod:


lib/plugins/rest.js
```diff
@@ -36,6 +37,76 @@ exports.register = function (server, options, next) {
...
+			/*Create a JSON object of all paramaters which are present in the GET Request*/
+			for (var key in request.query) {
+				if (key === "id") {
+					query.id = request.query.id;
+				}
+				else if (key === "type") {
+					query.type = request.query.type;
+				}
+				else if (key === "ami") {
+					query.key = request.query.key;
```
a-type: Is this a typo?

lib/plugins/rest.js
```diff
@@ -36,6 +37,76 @@ exports.register = function (server, options, next) {
...
+	server.route({
+		method  : "GET",
+		path    : "/v1/instances",
+
+		handler : function (request, reply) {
+
+			var	query = {};
+			var bCheckPresenceofInValidParameter = false;
+			/*Create a JSON object of all paramaters which are present in the GET Request*/
+			for (var key in request.query) {
```
a-type: I think this may benefit from using Lodash's [`_.pick` function](https://lodash.com/docs#pick).

lib/services/instanceAdapter.js
```diff
@@ -14,6 +14,14 @@ function InstanceAdapter (mapper) {
 		var instance = new Instance({ id : uuid.v4(), ami : ami, type : type });
 		return mapper.create(instance);
 	});
+
+	this.getInstance = Bluebird.method(function (query) {
+		return mapper.findOne(Instance,query);
+	});
+
+	this.getAllInstances = Bluebird.method(function (query) {
+		return mapper.find(Instance,query);
```
a-type: A space between parameters here (and line 19) would improve readability.

package.json
```diff
@@ -43,6 +43,7 @@
     "genesis": "^1.0.0",
     "joi": "^6.5.0",
     "mongodb": "^2.0.35",
-    "uuid": "^2.0.1"
+    "uuid": "^2.0.1",
+	"boom": "^2.8.0"
```
a-type: Needs a tab to satisfy OCD :)

test/unit/plugins/rest_spec.js
```diff
@@ -79,6 +82,8 @@ describe("The Rest plugin", function () {
 					expect(response.headers.location, "location").to.match(location);
 					expect(response.payload.ami, "ami").to.equal(VALID_AMI);
 					expect(response.payload.type, "type").to.equal(VALID_TYPE);
+					/*Required for GET Requests*/
+					VALID_INSTANCE_ID = response.payload.id;
```
a-type: Doing this starts to mix state between test cases. It's better practice to create any resources you need to test functionality within a single test block. So, when you want to test fetching an instance, have that test create its own instance before making the GET request.

test/unit/plugins/rest_spec.js
```diff
@@ -134,5 +139,237 @@ describe("The Rest plugin", function () {
 				expect(result.statusCode).to.equal(500);
 			});
 		});
+		describe("with an invalid instance", function () {
```
a-type: Our `describe` blocks are starting to get a little disorganized. Making the test output readable is helpful for debugging when things to wrong. I think our test output should look a little more like this:

```
The Rest plugin
    creating a new instance
        with valid parameters
        [ ... ]
    getting an instance
        with valid parameters
        [ ... ]
    querying for instances
        [ ... ]
```

This organizes the tests into specific, descriptive sections. You should be able to accomplish this by rearranging and adding `describe` blocks.

test/unit/plugins/rest_spec.js
```diff
@@ -134,5 +139,237 @@ describe("The Rest plugin", function () {
...
+					result = response;
+				});
+			});
+
+			it("returns a Bad Request", function () {
+				expect(result.statusCode,"status").to.equal(400);
+			});
+		});
+
+		describe("with a mix of valid and invalid parameters", function () {
```
a-type: Should this case result in a 400? I'm not sure we care if the user adds some extraneous parameters. In fact, there are certain cases where browsers or other middleware will add their own parameters (for instance, `nocache=true`) to a request for various reasons. Doing so shouldn't cause the request to fail if there are valid parameters present.

test/unit/plugins/rest_spec.js
```diff
@@ -134,5 +139,237 @@ describe("The Rest plugin", function () {
...
+
+			it("returns an empty array of intsances", function () {
+				var payload;
+				expect(result.statusCode,"status").to.equal(200);
+				payload = JSON.parse(result.payload);
+				expect(payload.instances.length).equal(0);
+			});
+		});
+	});
+	describe("when registered with stubbed instance", function () {
```
a-type: This description may be a little too specific to your test implementation. Descriptions work better if they are more... generally descriptive, instead of prescribing a certain way of doing things. For instance, `"encountering an internal error"` might work here, since these test cases are for 500 error situations. You accomplish the tests by stubbing, but that's not necessarily relevant to the test results output.

a-type:
Great.

Issue #59, opened 2015-06-29
Launch an EC2 instance when a new Instance is created


a-type:
Extend the adapter class which creates Instance resources to actually launch a real Amazon EC2 instance when a new Instance resource is created. This EC2 instance should use the specified AMI and instance type, passing them to the [Amazon AWS SDK](http://aws.amazon.com/sdk-for-node-js/). It should then update the `uri` field of the instance model with the location of the create instance before the request completes.

Test the functionality by stubbing the AWS SDK methods necessary to create new instances, and verify that they are being called with the correct parameters. Test relevant scenarios and cases:
- An instance is launched without specifying AMI or type, and defaults to the Ubuntu basic AMI and `t2.micro` instance type.
- An instance is launched with a specified AMI, and passes that AMI to the AWS SDK, along with the default instance type.
- An instance is launched with a specified instance type, and passes that type to the AWS SDK, along with the default AMI.
- An instance is launched with both a specified AMI and instance type, and passes those values to the AWS SDK.
- The AWS SDK encounters an error (make your stub return an error) while launching an instance.

Also test an end-to-end scenario using the REST API endpoints from #57 and #58:
- A POST is made to `/v1/instances`, and the reply body includes a `uri` property which was provided by an AWS SDK stub for testing purposes.

**NOTE:** when testing, don't ever hit AWS directly, or launch any real instances. Use stubbing and mocking to simulate the behavior you want to test against, whether that be functioning or malfunctioning behavior. Launching real EC2 instances from tests without cleaning up could be costly if we don't catch it!

PR #77, opened 2015-07-15
EC2 Instances can be created from AWS.

arjunmadan:
Added the AWS SDK - now lets EC2 instances get created.
Reworked some test cases for the Rest plugin and InstanceAdapter so that they are actually unit tests.


lib/plugins/rest.js
```diff
@@ -4,10 +4,21 @@ var InstanceAdapter = require("../services/instanceAdapter.js");
 var Boom            = require("boom");
 var _               = require("lodash");
 var Joi             = require("joi");
+var AwsAdapter      = require("../services/awsAdapter");
+var AWS             = require("aws-sdk");
+var Bluebird        = require("bluebird");
+
+var ec2 = new AWS.EC2({
```
a-type: Any reason to have this intialization logic run at the start, rather than in the `register` setup?

lib/plugins/rest.js
```diff
@@ -26,16 +37,27 @@ exports.register = function (server, options, next) {
...
 					reply("Bad Request: Please check the parameters passed.").code(400);
 				}
+				else if (error.name === "InvalidAMIID.Malformed") {
+					reply("The AMI entered does not exist. Ensure it is of the form ami-xxxxxx.").code(400);
+				}
+				else if (error.name === "InvalidParameterValue") {
+					reply("The Type entered does not exist. Ensure it is a valid EC2 type.").code(400);
+				}
+				else if (error.name === "AuthFailure") {
+					reply("Authentication Failure. Ensure your AWS credentials have been correctly used.").code(400);
```
a-type: This isn't actually a 400, since the credentials are configured in the server. The error here is due to the server admin not starting the server with the correct configuration, so it should be a 500 error.

lib/server.js
```diff
@@ -4,7 +4,7 @@ var Hapi           = require("hapi");
 var Bluebird       = require("bluebird");
 var MongoMapper    = require("genesis").MongoMapper;
 
-var mapper = new MongoMapper(process.env.MONGODB_URL);
+var mapper = new MongoMapper(process.env.MONGOLAB_URI);
```
a-type: Best practice is to configure your deployment environment to suit your code, rather than change your code to expect a specific deployment environment. Using the generic `MONGODB_URL` variable lets us switch between Heroku/AWS, or MongoLab/RDS/Compose, without having to modify and re-deploy our code. Instead, we can just point `MONGODB_URL` to whatever we have configured in our environment.

lib/services/awsAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+				},
+				{
+					Key   : "ID",
+					Value : instance.id
+				} ]
+			};
+			ec2.createTagsAsync(tags);
+			return instance;
+		})
+		.catch(function (error) {
```
a-type: This shouldn't be necessary-- if no `.catch` block exists, the error will continue to propogate to the next part of the chain (i.e., the function that is calling `runInstances`).

lib/services/awsAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+				Tags      : [ {
+					Key   : "Name",
+					Value : "Service Maker"
+				},
+				{
+					Key   : "ID",
+					Value : instance.id
+				} ]
+			};
+			ec2.createTagsAsync(tags);
```
a-type, line 40: Should we wait until the tags are created before returning the instance?

test/unit/plugins/rest_spec.js
```diff
@@ -68,20 +54,44 @@ describe("The Rest plugin", function () {
...
+
+		var awsAdapter = new AwsAdapter();
 
 		before(function () {
+			createInstanceStub = Sinon.stub(instanceAdapter, "createInstance")
+			.returns(Bluebird.resolve({
+				ami   : "ami-d05e75b8",
+				type  : "t2.micro",
+				state : "pending",
+				url   : ""
```
a-type: This should be `uri`.

test/unit/plugins/rest_spec.js
```diff
@@ -131,35 +137,129 @@ describe("The Rest plugin", function () {
...
+
+		var awsAdapter = new AwsAdapter();
+
+		before(function () {
+			createInstanceStub = Sinon.stub(instanceAdapter, "createInstance")
+			.returns(Bluebird.resolve({
+				ami   : "ami-d05e75b8",
+				type  : "t2.micro",
+				state : "pending",
+				url   : ""
```
a-type: This should be `uri`.

test/unit/plugins/rest_spec.js
```diff
@@ -131,35 +137,129 @@ describe("The Rest plugin", function () {
...
+				}
+			});
+		});
+
+		after(function () {
+			createInstanceStub.restore();
+			return server.stopAsync();
+		});
+
+		describe("with an ami that doesn't exist", function () {
```
a-type: This block appears to be describing an authentication failure.

test/unit/plugins/rest_spec.js
```diff
@@ -131,35 +137,129 @@ describe("The Rest plugin", function () {
...
+					expect(response.payload)
+					.to.equal("The Type entered does not exist. Ensure it is a valid EC2 type.");
+				});
 			});
 		});
+
 	});
 
-	describe("getting an instance", function () {
+	describe("when there is a problem with the database connection", function () {
```
a-type, line 272: I think this block should be moved into the `creating an instance` block above.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +49,181 @@ describe("The InstanceAdapter class ", function () {
 
 	});
 
-	describe("when the mapper fails to create a new model", function () {
-		var result;
-		var mapperStub;
+	describe("Getting an instance", function () {
```
a-type, line 52: Huh, did we not already have these tests?

a-type:
I'm going to assign this PR to Jeff for review tomorrow after comments are addressed / discussed.

lib/plugins/rest.js
```diff
@@ -22,20 +34,61 @@ exports.register = function (server, options, next) {
...
+						type  : result.type,
+						ami   : result.ami,
+						state : "failed",
+						uri   : null
+					});
+
+					instances.updateInstance(failedInstance);
+					reply("The Type entered does not exist. Ensure it is a valid EC2 type.").code(400);
+				}
+				else if (error.name === "AuthFailure") {
```
jagoda: Would a `switch` statement be a better choice for this branching structure since they all just check `error.name`?

jagoda: You may also want to consider the [Bluebird `.catch` handler](https://github.com/petkaantonov/bluebird/blob/master/API.md#catchfunction-errorclassfunction-predicate-function-handler---promise).

lib/plugins/rest.js
```diff
@@ -22,20 +34,61 @@ exports.register = function (server, options, next) {
...
+					failedInstance = new Instance({
+						id    : result.id,
+						type  : result.type,
+						ami   : result.ami,
+						state : "failed",
+						uri   : null
+					});
+
+					instances.updateInstance(failedInstance);
+					reply("The AMI entered does not exist. Ensure it is of the form ami-xxxxxx.").code(400);
```
jagoda: Ideally you should use [Boom](https://github.com/hapijs/boom) to generate HTTP errors.

lib/plugins/rest.js
```diff
@@ -22,20 +34,61 @@ exports.register = function (server, options, next) {
...
+					failedInstance = new Instance({
+						id    : result.id,
+						type  : result.type,
+						ami   : result.ami,
+						state : "failed",
+						uri   : null
+					});
+
+					instances.updateInstance(failedInstance);
+					reply("The Type entered does not exist. Ensure it is a valid EC2 type.").code(400);
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34893346).

lib/plugins/rest.js
```diff
@@ -22,20 +34,61 @@ exports.register = function (server, options, next) {
...
+					failedInstance = new Instance({
+						id    : result.id,
+						type  : result.type,
+						ami   : result.ami,
+						state : "failed",
+						uri   : null
+					});
+
+					instances.updateInstance(failedInstance);
+					reply("Authentication Failure. Ensure your AWS credentials have been correctly used.").code(500);
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34893346).

test/unit/plugins/rest_spec.js
```diff
@@ -1,44 +1,30 @@
...
-				response.payload = JSON.parse(response.payload);
-				expect(response.statusCode, "status").to.equal(201);
-				expect(response.headers.location, "location").to.match(location);
-				expect(response.payload.ami, "ami").to.equal(VALID_AMI);
-				expect(response.payload.type, "type").to.equal(VALID_TYPE);
-				VALID_INSTANCE_ID = response.payload.id;
-			});
-	}
+	//var DEFAULT_AMI       = "ami-d05e75b8";
+	//var DEFAULT_TYPE      = "t2.micro";
```
jagoda: We should remove dead code.

test/unit/plugins/rest_spec.js
```diff
@@ -68,25 +54,59 @@ describe("The Rest plugin", function () {
...
+				type  : "t2.micro",
+				state : "pending",
+				uri   : ""
+			}));
+
+			updateInstanceStub = Sinon.stub(instanceAdapter, "updateInstance")
+			.returns(Bluebird.resolve({
+				ami   : "ami-d05e75b8",
+				type  : "t2.micro",
+				state : "failed",
```
jagoda: Hard-coding the mock to always returned the `failed` status seems to gloss over the fact that we need the REST API to explicitly set this state. We need our test cases to demonstrate that this is done correctly.

test/unit/plugins/rest_spec.js
```diff
@@ -131,40 +147,180 @@ describe("The Rest plugin", function () {
 			});
 		});
 
-		describe("when there is a problem with the database connection", function () {
-			var result;
+	});
+
+	describe("creating a new instance", function () {
```
jagoda: How is this different that the `describe` block above (the description looks the same)?

test/unit/plugins/rest_spec.js
```diff
@@ -131,40 +147,180 @@ describe("The Rest plugin", function () {
...
+				type  : "t2.micro",
+				state : "pending",
+				uri   : ""
+			}));
+
+			updateInstanceStub = Sinon.stub(instanceAdapter, "updateInstance")
+			.returns(Bluebird.resolve({
+				ami   : "ami-d05e75b8",
+				type  : "t2.micro",
+				state : "failed",
```
jagoda, line 164: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34894202).

test/unit/plugins/rest_spec.js
```diff
@@ -131,40 +147,180 @@ describe("The Rest plugin", function () {
...
+			});
+
+			it("returns an error with statusCode 400", function () {
+				var request = new Request("POST", "/v1/instances").mime("application/json").payload({
+					ami  : VALID_AMI,
+					type : VALID_TYPE
 				});
+				return request.inject(server)
+				.then(function (response) {
+					expect(response.statusCode, "status").to.equal(500);
```
jagoda: The description says that the status code should be `400` while the test case checks for `500`. Which should it actually be?

lib/plugins/rest.js
```diff
@@ -22,20 +34,61 @@ exports.register = function (server, options, next) {
...
 			var type = request.payload.type;
 			instances.createInstance(ami, type)
 			.then(function (instance) {
+				result = instance;
+				return awsAdapter.runInstances(instance);
+			})
+			.then(function (instance) {
 				reply(instance).created("/v1/instances/" + instance.id);
 			})
 			.catch(function (error) {
```
jagoda, line 48: We probably want to make sure that we log these errors at some point.

test/unit/plugins/rest_spec.js
```diff
@@ -131,40 +147,180 @@ describe("The Rest plugin", function () {
...
+					}
 				});
 			});
 
 			after(function () {
 				mapper.create.restore();
+				return server.stopAsync();
 			});
 
 			it("displays an error page", function () {
```
jagoda: Is this description still accurate? This looks like it is for a REST API and not a web UI.

test/unit/plugins/rest_spec.js
```diff
@@ -175,22 +331,43 @@ describe("The Rest plugin", function () {
...
 		});
 
 		describe("with a valid instance", function () {
+			var getInstanceStub;
+
 			before(function () {
-				createAnInstance(server);
+				getInstanceStub = Sinon.stub(instanceAdapter, "getInstance")
+				.returns(Bluebird.resolve({
+					id    : VALID_INSTANCE_ID,
```
jagoda, line 356: We should probably check for this payload in the assertions below.

test/unit/plugins/rest_spec.js
```diff
@@ -222,9 +399,22 @@ describe("The Rest plugin", function () {
 		});
 
 		describe("with a valid type", function () {
+			var getAllInstancesStub;
 
 			before(function () {
-				createAnInstance(server);
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
+				.returns(Bluebird.resolve([ {
+						id    : VALID_INSTANCE_ID,
```
jagoda, line 408: We should probably check for this payload in the assertions below.

jagoda, line 408: I don't think that this test is validating what it claims to be validating. Because the `getAllInstances()` method is stubbed and hard-coded to return a fixed instance list and the test case never validates that the correct type argument is sent to the method, we have no idea that the query is in fact processing correctly.

test/unit/plugins/rest_spec.js
```diff
@@ -240,8 +430,12 @@ describe("The Rest plugin", function () {
 
 		describe("with an invalid instance by sending non-exisitng Instance Id and type", function () {
 			var result;
+			var getAllInstancesStub;
 
 			before(function () {
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
+				.returns(Bluebird.resolve([ ]));
```
jagoda, line 442: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34895388).

test/unit/plugins/rest_spec.js
```diff
@@ -259,8 +457,19 @@ describe("The Rest plugin", function () {
 
 		describe("with a valid ID", function () {
 			var result;
+			var getAllInstancesStub;
 
 			before(function () {
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
```
jagoda, line 472: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34895388).

test/unit/plugins/rest_spec.js
```diff
@@ -279,8 +492,19 @@ describe("The Rest plugin", function () {
 
 		describe("with a valid AMI", function () {
 			var result;
+			var getAllInstancesStub;
 
 			before(function () {
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
```
jagoda, line 510: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34895388).

test/unit/plugins/rest_spec.js
```diff
@@ -295,8 +523,19 @@ describe("The Rest plugin", function () {
 
 		describe("with state=pending", function () {
 			var result;
+			var getAllInstancesStub;
 
 			before(function () {
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
```
jagoda, line 543: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34895388).

test/unit/plugins/rest_spec.js
```diff
@@ -311,8 +554,12 @@ describe("The Rest plugin", function () {
 
 		describe("with invalid URI", function () {
 			var result;
+			var getAllInstancesStub;
 
 			before(function () {
+				getAllInstancesStub = Sinon.stub(instanceAdapter, "getAllInstances")
```
jagoda, line 576: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34895388).

test/unit/services/awsAdapter_spec.js
```diff
@@ -0,0 +1,174 @@
...
+	var VALID_INSTANCE = {
+		ami  : DEFAULT_IMAGE_ID,
+		type : DEFAULT_INSTANCE_TYPE
+	};
+
+	var INVALID_INSTANCE = {
+		ami  : INVALID_IMAGE_ID,
+		type : DEFAULT_INSTANCE_TYPE
+	};
+	describe("creating a new instance with valid ami, type", function () {
```
jagoda: This test case (like several others in this PR) is overly reliant on stub implementations. This test does not actually verify that the AWS adapter correctly creates an instance since it does not check that the AWS SDK is invoked with the expected arguments. This essentially makes this test case a tautology -- it passes because it the stubs return what we told them to.

test/unit/services/awsAdapter_spec.js
```diff
@@ -0,0 +1,174 @@
...
+			createTagsStub.restore();
+		});
+
+		it("returns a new instance with the ami and type provided", function () {
+			expect(result.ami, "response").to.equal("ami-d05e75b8");
+			expect(result.type, "response").to.equal("t2.micro");
+		});
+	});
+
+	describe("creating a new instance with invalid ami, type", function () {
```
jagoda: There are three different `describe` blocks with this same title. We should either group all of the assertions (`it` blocks) into the same `describe` block because they are testing the same scenario (as the description implies) or clarify the description to indicate what's different.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -22,11 +22,29 @@ describe("The InstanceAdapter class ", function () {
 
 			instances.createInstance()
 			.then(function (result) {
+				var failedInstance = new Instance({
```
jagoda: This seems extraneous. Can you explain why it is used here?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -22,11 +22,29 @@ describe("The InstanceAdapter class ", function () {
...
+					type  : result.type,
+					uri   : result.uri
+				});
+
 				expect(result).to.be.an.instanceOf(Instance);
 				expect(result.id).to.match(ID_REGEX);
 				expect(result.type).to.equal(DEFAULT_TYPE);
 				expect(result.ami).to.equal(DEFAULT_AMI);
+
+				return instances.updateInstance(failedInstance)
```
jagoda: What is the rationale for calling `updateInstance()` in the test case for creating instances?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
+			var ID_REGEX     = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+			var instances    = new InstanceAdapter();
+
+			before(function () {
+				instances.createInstance()
+				.then(function (response) {
+					expect(response).to.be.an.instanceOf(Instance);
+					expect(response.id).to.match(ID_REGEX);
+					expect(response.type).to.equal(DEFAULT_TYPE);
+					expect(response.ami).to.equal(DEFAULT_AMI);
```
jagoda: What is the thought behind having these assertions here? If the `createInstance()` function works correctly, can't we just assume that this information is correct?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
+				.then(function (result) {
+					expect(result).to.be.an.instanceOf(Instance);
+					expect(result.id).to.match(ID_REGEX);
+					expect(result.type).to.equal(DEFAULT_TYPE);
+					expect(result.ami).to.equal(DEFAULT_AMI);
+				});
+			});
+		});
+
+		describe("with an invalid instanceId", function () {
```
jagoda: Is this an 'invalid' instance ID or a non-existent instance ID?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
-			var instances = new InstanceAdapter(mapper);
 
-			mapperStub = Sinon.stub(mapper, "create")
-				.rejects(new Error("Simulated Failure"));
+			instances.createInstance()
+			.then(function (response) {
+				expect(response).to.be.an.instanceOf(Instance);
+				expect(response.id).to.match(ID_REGEX);
+				expect(response.type).to.equal(DEFAULT_TYPE);
+				expect(response.ami).to.equal(DEFAULT_AMI);
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/77/files#r34897266).

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
 
-			return instances.createInstance(DEFAULT_AMI, DEFAULT_TYPE)
-			.catch(function (err) {
-				result = err;
+		describe("with valid parameters", function () {
+			it("receives an array of instances", function () {
+				instances.getAllInstances({ ami : DEFAULT_AMI, type : DEFAULT_TYPE })
+				.then(function (response) {
+					expect(response).to.be.an.instanceOf(Array);
+					expect(response.length).to.be.at.least(1);
```
jagoda: Shouldn't we also make sure that we returned the right object?

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
 		});
 
-		after(function () {
-			mapperStub.restore();
+		describe("with no parameters", function () {
+			it("receives an array of all instances", function () {
+				instances.getAllInstances()
+				.then(function (response) {
+					expect(response).to.be.an.instanceOf(Array);
+					expect(response.length).to.be.at.least(1);
```
jagoda: This test is functionally no different that the previous one. How do we know that anything different happened? In other words, this test case doesn't actually prove that the query returns all of the instances and not just some arbitrary first instance.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -49,33 +67,181 @@ describe("The InstanceAdapter class ", function () {
...
+				});
+			});
 		});
 
-		it("throws an error", function () {
-			expect(result, "error").to.be.an.instanceOf(Error);
-			expect(result.message, "error message").to.equal("Simulated Failure");
+		describe("with invalid parameters", function () {
```
jagoda: I think that this description is misleading. 'Invalid' makes me think that the parameter is malformed or not allowed. This test case seems to be checking what happens if the query doesn't match any instances.

jagoda:
The functional code is looking pretty good. However, the test cases still need some work. It may be helpful to think about test cases as formal proofs of correctness for the functional code and not just as assertions that a specific result was received as the result of a function call.

jagoda:
Closes #59.

test/unit/services/instanceAdapter_spec.js
arjunmadan: This was used to check whether `updateInstance()` is called to change the `state` to failed in case the AWS SDK fails.

test/unit/services/instanceAdapter_spec.js
arjunmadan: This was to check whether AWS failed. I've reworded the test cases to make them clearer.

jagoda: Hmmm... OK. I'm assuming you haven't pushed those changes yet. Correct?

test/unit/services/instanceAdapter_spec.js
jagoda: I still don't understand why this is needed for the simple create test case. Perhaps it will be more clear with the changes you reference below?

test/unit/services/instanceAdapter_spec.js
arjunmadan: Yes, I will have something pushed soon.

jagoda: Awesome.

jagoda:
Looks good. Thanks for all the hard work on this!

PR #83, opened 2015-07-22
Fixing stubbing issue in rest_spec

arjunmadan:


.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: node_js
 node_js:
-  - "0.12"
+- "0.12"
 branches:
   only:
   - v1.0
+deploy:
```
jagoda, line 7: This seems unrelated to the test cleanup.

jagoda:
There seem to be a lot of changes listed here. It would be nice if this change was made in it's own branch so that only the relevant commits appear in the pull request.

.travis.yml
arjunmadan, line 7: This was when I was trying to setup Travis to deploy automatically to heroku. Should I rename the PR to mention this as well?

PR #85, opened 2015-07-22
Issue60 merge

rrpod:


lib/plugins/rest.js
```diff
@@ -43,7 +45,34 @@ exports.register = function (server, options, next) {
 				return awsAdapter.runInstances(instance);
 			})
 			.then(function (instance) {
+				var updatedInstance;
+
 				reply(instance).created("/v1/instances/" + instance.id);
+				awsInstanceID = instance.instanceID;
```
jagoda: It looks like this variable isn't used outside of this block. Is there a reason that it needs to be scoped at the handler level?

lib/plugins/rest.js
```diff
@@ -43,7 +45,34 @@ exports.register = function (server, options, next) {
...
+				return sshAdapter.startSshPolling(awsInstanceID)
+				.then(function () {
+					return awsAdapter.getPublicIPAddress(awsInstanceID);
+				})
```
jagoda: I know we talked about this in person and I didn't have an answer at the time. Now that I see this code though, I wonder if we will ever want an instance and not want SSH enabled and the public IP available. This leads me to think that these function calls should be internal to the `AwsAdapter` and should be called implicitly before returning from `runInstances()`.

lib/server.js
```diff
@@ -4,7 +4,7 @@ var Hapi           = require("hapi");
 var Bluebird       = require("bluebird");
 var MongoMapper    = require("genesis").MongoMapper;
 
-var mapper = new MongoMapper(process.env.MONGODB_URL);
+var mapper = new MongoMapper(process.env.MONGODB_URI);
```
jagoda: This seems unrelated to this PR. Is there a reason that this is changing in the PR?

lib/services/awsAdapter.js
```diff
@@ -44,6 +43,28 @@ function AWSAdapter (ec2) {
...
+		};
+		return ec2.describeInstancesAsync(params)
+		.then(function (data) {
+			var instanceInfo = [];
+			data.Reservations.forEach(function (reservation) {
+				reservation.Instances.forEach(function (instance) {
+					instanceInfo.push(instance);
+				});
+			});
+			return Bluebird.resolve(instanceInfo);
```
jagoda: The `Bluebird.resolve()` call here is unnecessary. If you simply return the value then Bluebird will use it as the promise resolution automatically.

lib/services/awsAdapter.js
```diff
@@ -44,6 +43,28 @@ function AWSAdapter (ec2) {
 			return result;
 		});
 	});
+
+	this.getPublicIPAddress = Bluebird.method(function (instanceID) {
```
jagoda: I believe that the use of `Bluebird.method()` here is unnecessary since the method already does all of it's work in a promise chain.

lib/services/awsAdapter.js
```diff
@@ -44,6 +43,28 @@ function AWSAdapter (ec2) {
...
+		return this.describeInstance(instanceID)
+		.then(function (instance) {
+			return	Bluebird.resolve(instance[ 0 ].PublicIpAddress);
+		});
+	});
+	this.describeInstance = Bluebird.method(function (instanceID) {
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243112).

lib/services/instanceAdapter.js
```diff
@@ -1,9 +1,9 @@
 "use strict";
 
-var Instance = require("../models/Instance");
-var Genesis  = require("genesis");
-var uuid     = require("uuid");
-var Bluebird = require("bluebird");
+var Instance   = require("../models/Instance");
+var Genesis    = require("genesis");
+var uuid       = require("uuid");
+var Bluebird   = require("bluebird");
```
jagoda: ?

lib/services/instanceAdapter.js
```diff
@@ -11,7 +11,7 @@ function InstanceAdapter (mapper) {
 	}
 
 	this.createInstance = Bluebird.method(function (ami, type) {
-		var instance = new Instance({ id : uuid.v4(), ami : ami, type : type });
+		var instance   = new Instance({ id : uuid.v4(), ami : ami, type : type });
```
jagoda: ?

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+
+var Bluebird = require("bluebird");
+
+function SshAdapter (ec2) {
+	var SSH_POLLING_INTERVAL = 30; /*Time in Seconds*/
+
+	this.checkIfSshable = Bluebird.method(function (instanceID) {
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243112).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+		return this.canSsh(instanceID)
+		.then(function (sshable) {
+			if (sshable) {
+				return Bluebird.resolve();
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243012).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+			}
+			else {
+				return Bluebird.delay(SSH_POLLING_INTERVAL * 1000)
+				.then(this.checkIfSshable.bind(this, instanceID));
+			}
+		}.bind(this));
```
jagoda: While this is a valid thing to do, I find that it tends to make the code more difficult to read since it changes the normal receiver of the function and does so in a way that is not localized to the function body. I prefer using something like `var self = this;` in the outer scope and then using `self` rather than `this` throughout the function body.

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	this.startSshPolling = Bluebird.method(function (instanceID) {
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243112).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+
+		return this.checkIfSshable(instanceID)
+		.catch(function (err) {
+			return Bluebird.reject(err);
+		});
+	});
+
+	this.canSsh = Bluebird.method(function (instanceID) {
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243112).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+				.then(this.checkIfSshable.bind(this, instanceID));
+			}
+		}.bind(this));
+	});
+
+	this.startSshPolling = Bluebird.method(function (instanceID) {
+
+		return this.checkIfSshable(instanceID)
+		.catch(function (err) {
+			return Bluebird.reject(err);
```
jagoda: The `Bluebird.reject()` call here is unnecessary since this code is internal to a promise chain. You can simply throw the error. Furthermore, this block has no effect since it simply catches and re-throws an error without changing anything. We should probably just remove this frame.

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+
+	this.canSsh = Bluebird.method(function (instanceID) {
+		var sshable;
+		var params = {
+			InstanceIds : [ instanceID ]
+		};
+
+		return ec2.describeInstanceStatusAsync(params)
+		.then(function (data) {
+			sshable = false;
```
jagoda: It looks like this variable is only used in this promise frame. Is there a reason that it is scoped to the entire method?

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+			if (data.InstanceStatuses.length === 0) {
+				return (Bluebird.resolve(sshable));
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243012).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+
+					/* falls through */
+					case "insufficient-data" : {
+						return Bluebird.reject(new Error("System Checks Failed"));
+					}
+				}
+				return (Bluebird.resolve(sshable));
+			}
+		})
+		.catch(function (err) {
```
jagoda: This block has no effect. See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35243912).

test/unit/models/instance_spec.js
```diff
@@ -154,4 +154,4 @@ describe("The Instance model", function () {
 		true
 	);
 
-});
+});
```
jagoda: It looks like you didn't make any changes to this file. You should consider using an editor with [EditorConfig](http://editorconfig.org/) support so that this sort of thing is handled automatically (the project has an `.editorconfig` file).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,64 @@
...
+			return Bluebird.reject(err);
+		});
+	});
+}
+Object.freeze(SshAdapter);
+
+module.exports = SshAdapter;
```
jagoda: See [EditorConfig comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35244477) below.

test/unit/plugins/rest_spec.js
```diff
@@ -15,15 +16,16 @@ require("sinon-as-promised");
...
 
 describe("The Rest plugin", function () {
-	var VALID_INSTANCE_ID = "da14fbf2-5404-4f92-b55f-a961578204ed";
-	var VALID_AMI         = "ami-d05e75b8";
-	var VALID_TYPE        = "t2.micro";
-	var ID_REGEX          = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
+	var VALID_INSTANCE_ID  = "da14fbf2-5404-4f92-b55f-a961578204ed";
+	var VALID_AMI          = "ami-d05e75b8";
+	var VALID_TYPE         = "t2.micro";
+	var ID_REGEX           = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
```
jagoda: ?

test/unit/plugins/rest_spec.js
```diff
@@ -51,18 +53,22 @@ describe("The Rest plugin", function () {
 			});
 		});
 	});
-
```
jagoda, line 56: I think the extra line of whitespace here is a good thing (just my opinion). Was it removed intentionally?

test/unit/plugins/rest_spec.js
```diff
@@ -105,11 +133,32 @@ describe("The Rest plugin", function () {
 					response.payload = JSON.parse(response.payload);
 					expect(response.statusCode, "status").to.equal(201);
 					expect(response.headers.location, "location").to.match(location);
+					expect(startSshPollingStub.called).to.be.true;
+					expect(getPublicIPAddressStub.called).to.be.true;
+					expect(updateInstanceStub.called).to.be.true;
```
jagoda: Do we care what arguments are passed to these methods?

test/unit/plugins/rest_spec.js
```diff
@@ -95,6 +104,25 @@ describe("The Rest plugin", function () {
...
+
+				getPublicIPAddressStub = Sinon.stub(awsAdapter, "getPublicIPAddress", function () {
+					return (Bluebird.resolve(VALID_IP_ADDRESS));
+				});
+			});
+
+			after(function () {
+				getPublicIPAddressStub.restore();
+				startSshPollingStub.restore();
+				updateInstanceStub.restore();
```
jagoda: Sharing stubs like this has a tendency to create false positives. It is very easy to forget to call `.restore()` and to not know that you are working with state from a previous test case. It would be better if we could find a way to rework these test cases so that they don't share state.

test/unit/plugins/rest_spec.js
```diff
@@ -121,24 +170,114 @@ describe("The Rest plugin", function () {
...
 				});
 				return request.inject(server)
 				.then(function (response) {
-					var error = JSON.parse(response.payload);
-					expect(error.statusCode, "status").to.equal(400);
-					expect(error.message)
-					.to.equal("Bad Request: Please check the parameters passed.");
+					response.payload = JSON.parse(response.payload);
+					expect(response.statusCode, "status").to.equal(201);
+					expect(response.headers.location, "location").to.match(location);
```
jagoda: This test case doesn't look any different than the success case. What exactly do we expect to happen here that's different?

test/unit/plugins/rest_spec.js
```diff
@@ -121,24 +170,114 @@ describe("The Rest plugin", function () {
...
+
+			it("fails", function () {
+				var request = new Request("POST", "/v1/instances").mime("application/json").payload({
+					ami  : VALID_AMI,
+					type : VALID_TYPE
+				});
+				return request.inject(server)
+				.then(function (response) {
+					response.payload = JSON.parse(response.payload);
+					expect(response.statusCode, "status").to.equal(201);
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35245659).

test/unit/plugins/rest_spec.js
```diff
@@ -266,42 +405,41 @@ describe("The Rest plugin", function () {
...
-					type : VALID_TYPE
-				});
-				return request.inject(server)
-				.then(function (response) {
-					expect(response.result.message).to.equal("An internal server error occurred");
-					expect(response.statusCode).to.equal(500);
-				});
+		after(function () {
+			mapper.create.restore();
+			return server.stopAsync();
```
jagoda: Is the server ever started?

test/unit/plugins/rest_spec.js
```diff
@@ -266,42 +405,41 @@ describe("The Rest plugin", function () {
...
-				Sinon.stub(mapper, "create").rejects(new Error("Simulated Failure."));
-				server.connection();
-				return server.registerAsync({
-					register : Rest,
-					options  : {
-						mapper : mapper
-					}
-				});
-			});
+		var mapper = new MemoryMapper();
```
jagoda: I don't understand what changed in these test cases. This lines look unrelated to the SSH functionality.

test/unit/plugins/rest_spec.js
```diff
@@ -121,24 +170,114 @@ describe("The Rest plugin", function () {
...
+				.then(function (response) {
+					response.payload = JSON.parse(response.payload);
+					expect(response.statusCode, "status").to.equal(201);
+					expect(response.headers.location, "location").to.match(location);
+				});
+			});
+		});
 	});
 
+	describe("with invalid parameter(s) passed", function () {
```
jagoda: Should this test case have been part of a separate PR? It seems unrelated to the SSH feature.

test/unit/plugins/rest_spec.js
```diff
@@ -419,6 +557,7 @@ describe("The Rest plugin", function () {
 			});
 
 			it("returns an array of instances with the given type", function () {
+
```
jagoda: ?

test/unit/plugins/rest_spec.js
```diff
@@ -460,7 +599,6 @@ describe("The Rest plugin", function () {
 				expect(getAllInstancesStub.args[ 0 ][ 0 ].type).to.equal(INVALID_QUERY);
 				expect(getAllInstancesStub.args[ 0 ][ 0 ].id).to.equal(INVALID_QUERY);
 				expect(Object.keys(getAllInstancesStub.args[ 0 ][ 0 ]).length).to.equal(2);
-
```
jagoda, line 435: ?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+	describe("gets properties of given instance", function () {
+		var describeInstancesStub;
+		var result;
+		before(function () {
+			var awsAdapter = new AwsAdapter(ec2);
+			describeInstancesStub = Sinon.stub(ec2, "describeInstancesAsync", function () {
+				var data = { Reservations : [ { Instances : [ { PublicIpAddress : "127.0.0.1" } ] } ] };
+				return Bluebird.resolve(data);
+			});
+			awsAdapter.describeInstance(VALID_EC2_INSTANCE)
```
jagoda: I think this before block is prone to timing errors since the promise chain is never returned.

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+			});
+		});
+
+		after(function () {
+			describeInstancesStub.restore();
+		});
+
+		it("gets properties of the given instance", function () {
+			expect(result).to.be.instanceof(Array);
+			expect(result[ 0 ].PublicIpAddress).to.be.equal("127.0.0.1");
```
jagoda, line 189: Do we care if there is more than one element in the resulting array?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+
+	describe("gets properties of invalid instance", function () {
+		var describeInstancesStub;
+		var result;
+		before(function () {
+			var awsAdapter = new AwsAdapter(ec2);
+			describeInstancesStub = Sinon.stub(ec2, "describeInstancesAsync", function () {
+				return Bluebird.reject(new Error("Instance not Found"));
+			});
+			awsAdapter.describeInstance(VALID_EC2_INSTANCE)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+			.then(function (response) {
+				result = response;
```
jagoda: This seems unnecessary since we never use the `response` value.

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+				result = error;
+			});
+		});
+
+		after(function () {
+			describeInstancesStub.restore();
+		});
+
+		it("fails", function () {
+			expect(result,"error").to.be.an.instanceof(Error);
```
jagoda: Is it worth inspecting the attributes of the error to make sure that it is the error object that we expect?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+		var describeInstancesStub;
+
+		before(function () {
+			var awsAdapter = new AwsAdapter(ec2);
+			describeInstancesStub = Sinon.stub(ec2, "describeInstancesAsync", function () {
+				var data = { Reservations : [ { Instances : [ { PublicIpAddress : "127.0.0.1" } ] } ] };
+				return Bluebird.resolve(data);
+			});
+
+			awsAdapter.getPublicIPAddress(VALID_EC2_INSTANCE)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+			describeInstancesStub = Sinon.stub(ec2, "describeInstancesAsync", function () {
+				var data = { Reservations : [ { Instances : [ { PublicIpAddress : "127.0.0.1" } ] } ] };
+				return Bluebird.resolve(data);
+			});
+			awsAdapter.describeInstance(VALID_EC2_INSTANCE)
+			.then(function (response) {
+				result = response;
+			})
+			.catch(function (error) {
+				result = error;
```
jagoda: This seems unnecessary since we never use the error value.

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +150,88 @@ describe("The AwsAdapter class ", function () {
...
+				var data = { Reservations : [ { Instances : [ { PublicIpAddress : "127.0.0.1" } ] } ] };
+				return Bluebird.resolve(data);
+			});
+
+			awsAdapter.getPublicIPAddress(VALID_EC2_INSTANCE)
+			.then(function (response) {
+				result = response;
+			})
+			.catch(function (error) {
+				result = error;
```
jagoda, line 234: This seems unnecessary since we never use the error value.

test/unit/services/instanceAdapter_spec.js
```diff
@@ -217,32 +217,4 @@ describe("The InstanceAdapter class ", function () {
 			});
 		});
 	});
-
-	describe("updating an already created instance", function () {
```
jagoda, line 221: Can you explain why this test case is being removed? It seems unrelated to the SSH changes and removing test cases is usually not a great idea.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
+"use strict";
+
+var AWS        = require("aws-sdk");
+var sinon      = require("sinon");
+var expect     = require("chai").expect;
+var SshAdapter = require("../../../lib/services/sshAdapter");
+var Bluebird   = require("bluebird");
+var options    = {};
+var ec2        = new AWS.EC2();
```
jagoda: Is there any reason that the `options` and the `ec2` instance shouldn't be in the `describe` block? I don't think they will ever get garbage collected if they stay out here.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+					} ]
+				};
+				return (Bluebird.resolve(data));
+			});
+			sshAdapter.canSsh(id)
+			.then(function (data) {
+				result = data;
+			})
+			.catch(function (err) {
+				result = err;
```
jagoda: This looks like it is never used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			});
+		});
+
+		after(function () {
+			instanceStatusStub.restore();
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35245456).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+var ec2        = new AWS.EC2();
+
+Bluebird.promisifyAll(ec2);
+options.ec2 =  ec2;
+
+describe("The SSH Adapter Class", function () {
+	var id                   = "i-fbf47257";
+	var SSH_POLLING_INTERVAL = 30; //Time in Seconds
+
+	describe("Can Ssh", function () {
```
jagoda: I don't understand this description.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+
+		it("is able to connect", function () {
+			expect(instanceStatusStub.called).to.be.true;
+			expect(instanceStatusStub.args[ 0 ][ 0 ].InstanceIds[ 0 ]).to.equal(id);
+			expect(result).to.be.true;
+		});
+	});
+
+	describe("Tries to ssh", function () {
+		describe("as soon as instance is created", function () {
```
jagoda, line 52: I don't understand these descriptions. Ideally, these descriptions will align with each other so that when a test case fails it reads almost like a user story. So something like "The SSH adapter when trying to connect to ssh before the instance has started fails to connect.".

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				.then(function (data) {
+					result = data;
+				})
+				.catch(function (err) {
+					result = err;
+				});
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
```
jagoda, line 70: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35245456).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+						InstanceStatuses : []
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
+				.then(function (data) {
+					result = data;
+				})
+				.catch(function (err) {
+					result = err;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				});
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
+			});
+
+			it("it returns an error as system checks have failed", function () {
```
jagoda: It is probably sufficient to just use "it returns an error" here.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+									Status : "failed"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
+				.then(function (data) {
+					result = data;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+						SystemStatus : {
+							Details : [ {
+								Status : "passed"
+							} ]
+						}
+					} ]
+				};
+				return (Bluebird.resolve(data));
+			});
+			sshAdapter.canSsh(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			var result;
+			var sshAdapter = new SshAdapter(ec2);
+			before(function () {
+				instanceStatusStub = sinon.stub(ec2,"describeInstanceStatusAsync", function () {
+					var data = {
+						InstanceStatuses : []
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+							SystemStatus : {
+								Details : [ {
+									Status : "failed"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+							SystemStatus : {
+								Details : [ {
+									Status : "insufficient-data"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				.catch(function (err) {
+					result = err;
+				});
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
+			});
+
+			it("it returns an error as there is insufficient-data", function () {
```
jagoda: It is probably sufficient to just use "it returns an error" here.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+									Status : "insufficient-data"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+				sshAdapter.canSsh(id)
+				.then(function (data) {
+					result = data;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+							SystemStatus : {
+								Details : [ {
+									Status : "passed"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
```
jagoda: This `.tick()` call seems unnecessary. Am I missing something?

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+
+				sshAdapter.startSshPolling(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				.then(function (data) {
+					result = data;
+				})
+				.catch(function (err) {
+					result = err;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			before(function () {
+				clock = sinon.useFakeTimers();
+				instanceStatusStub = sinon.stub(ec2, "describeInstanceStatusAsync", function () {
+					var data = {
+						InstanceStatuses : []
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
```
jagoda: What does this do?

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
+				clock.restore();
+			});
+
+			it("is not able to connect", function () {
+				expect(instanceStatusStub.called).to.be.true;
+				expect(instanceStatusStub.args[ 0 ][ 0 ].InstanceIds[ 0 ]).to.equal(id);
```
jagoda: How do we know that it is not able to connect? It seems like we only check the arguments here and not any kind of resulting state.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+								Details : [ {
+									Status : "failed"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
```
jagoda: What is this for?

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				});
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
+				clock.restore();
+			});
+
+			it("is not able to connect as instance has failed", function () {
```
jagoda: It is probably sufficient to just say "is not able to connect" here.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+				sshAdapter.startSshPolling(id)
+				.then(function (data) {
+					result = data;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+									Status : "failed"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+				sshAdapter.startSshPolling(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+									Status : "insufficient-data"
+								} ]
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+				sshAdapter.startSshPolling(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+				});
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+			});
+
+			after(function () {
+				instanceStatusStub.restore();
+				clock.restore();
+			});
+
+			it("is not able to connect as system checks have failed", function () {
```
jagoda: It is probably sufficient to just use "it is not able to connect" here.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+							}
+						} ]
+					};
+					return (Bluebird.resolve(data));
+				});
+
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
+				sshAdapter.startSshPolling(id)
+				.then(function (data) {
+					result = data;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+		var instanceStatusStub;
+		var clock;
+		var result;
+		var sshAdapter = new SshAdapter(ec2);
+		before(function () {
+			clock = sinon.useFakeTimers();
+			instanceStatusStub = sinon.stub(ec2, "describeInstanceStatusAsync")
+				.rejects(new Error("Simulated Failure"));
+
+			clock.tick(SSH_POLLING_INTERVAL * 1000);
```
jagoda: What is this for?

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			sshAdapter.startSshPolling(id)
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35246272).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+			.then(function (data) {
+				result = data;
```
jagoda: I don't think that this is ever used.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,357 @@
...
+var expect     = require("chai").expect;
+var SshAdapter = require("../../../lib/services/sshAdapter");
+var Bluebird   = require("bluebird");
+var options    = {};
+var ec2        = new AWS.EC2();
+
+Bluebird.promisifyAll(ec2);
+options.ec2 =  ec2;
+
+describe("The SSH Adapter Class", function () {
```
jagoda, line 9: This test suite seems to only verify the behavior of the initial connection attempt. Unless I missed something, I didn't see any assertions validating that polling continues when SSH is not yet ready. Should we also support and validate a timeout if the instance does not fail but SSH also does not connect within a certain amount of time?

lib/plugins/rest.js
```diff
@@ -40,7 +40,7 @@ exports.register = function (server, options, next) {
 			instances.createInstance(ami, type)
 			.then(function (instance) {
 				result = instance;
-				return awsAdapter.runInstances(instance);
+				return awsAdapter.runInstances(instance, instances, server.log.bind(server));
```
jagoda: Would it make more sense to pass the `InstanceAdapter` to the `AwsAdapter` constructor?

jagoda: Same for the log function. I think these artifacts may end up getting used across several methods.

lib/services/awsAdapter.js
```diff
@@ -37,13 +46,74 @@ function AWSAdapter (ec2) {
...
+			var updatedInstance;
+
+			updatedInstance = new Instance({
+				id    : result.id,
+				type  : result.type,
+				ami   : result.ami,
+				state : "ready",
+				uri   : "https://" + ipAddress,
+			});
+			instances.updateInstance(updatedInstance);
```
jagoda: What happens if this fails? Will we know?

lib/services/awsAdapter.js
```diff
@@ -37,13 +46,74 @@ function AWSAdapter (ec2) {
...
+		.catch(function (error) {
+			var updatedInstance;
+			updatedInstance = new Instance({
+				id    : result.id,
+				type  : result.type,
+				ami   : result.ami,
+				state : "failed",
+				uri   : null
+			});
+			instances.updateInstance(updatedInstance);
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35444052).

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,58 @@
...
+		return this.canSsh(instanceID)
+		.then(function (sshable) {
+			if (sshable) {
+				return sshable;
+			}
+			else {
+				return Bluebird.delay(SSH_POLLING_INTERVAL * 1000)
+				.then(self.checkIfSshable.bind(self, instanceID));
+			}
+		}.bind(self));
```
jagoda: Would be great to not use bind here for the sake of readability.

test/unit/services/awsAdapter_spec.js
jagoda, line 189: Did this get missed?

test/unit/services/awsAdapter_spec.js
jagoda: ?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +156,166 @@ describe("The AwsAdapter class ", function () {
...
+					instanceProp  = JSON.parse(JSON.stringify(data));
+				});
+			});
+
+			after(function () {
+				startSshPollingStub.restore();
+				getPublicIPAddressStub.restore();
+			});
+
+			it("updates the state and uri of the instance", function () {
```
jagoda, line 287: This seems to only check one side of this scenario. I believe this test will pass even if `updateInstance()` is never called.

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +156,166 @@ describe("The AwsAdapter class ", function () {
...
+				.then(function (data) {
+					instanceProp  = JSON.parse(JSON.stringify(data));
+				});
+			});
+
+			after(function () {
+				startSshPollingStub.restore();
+			});
+
+			it("updates the state and uri of the instance", function () {
```
jagoda, line 362: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/85/files#r35444910).

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,285 @@
+"use strict";
+/*jslint plusplus: true */
```
jagoda: Is this really necessary? Seems like either this is wrong or the style rule is wrong.

test/unit/services/sshAdapter_spec.js
```diff
@@ -0,0 +1,285 @@
...
+			});
+
+			after(function () {
+				canSshStub.restore();
+				clock.restore();
+				delayStub.restore();
+			});
+
+			it("is possible to ssh eventually", function () {
+				clock.tick(SSH_POLLING_INTERVAL * 1000);
```
jagoda: Why is the clock advanced before anything happens?

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,58 @@
...
+	};
+
+	this.startSshPolling = function (instanceID) {
```
jagoda: This looks like it does exactly the same thing as `checkIfSshable()`. Do we really need both?

lib/services/sshAdapter.js
```diff
@@ -0,0 +1,58 @@
...
+	this.checkIfSshable = function (instanceID) {
+		var self = this;
+		return this.canSsh(instanceID)
+		.then(function (sshable) {
+			if (sshable) {
+				return sshable;
+			}
+			else {
+				return Bluebird.delay(SSH_POLLING_INTERVAL * 1000)
+				.then(self.checkIfSshable.bind(self, instanceID));
```
jagoda: The bind is unnecessary when using the `self` variable here.

jagoda:
We still need to make sure that SSH polling times out if it takes too long. Also, do you happen to know why Arjun's changes are also listed here? Did you merge his changes into your branch? It would be great if pull requests did not include outside changes.

test/unit/services/awsAdapter_spec.js
rrpod: Not sure how this came up. It isn't there in the current file. https://github.com/rrpod/service-maker/blob/issue60_merge/test/unit/services/awsAdapter_spec.js

test/unit/services/sshAdapter_spec.js
rrpod: I had tried without that initially. I got an error  Unexpected use of '++'. The jshint.json file already has plusplus set to true though

test/unit/services/awsAdapter_spec.js
jagoda: Huh. Ok, fair enough. I guess we'll see what happens when you push your latest code.

lib/services/awsAdapter.js
```diff
@@ -1,13 +1,47 @@
...
+		}
+		else if (options.ec2) {
+			sshAdapter = new SshAdapter(options.ec2);
+		}
+		instances  = options.instances;
+		serverLog  = options.serverLog;
+		ec2        = options.ec2;
+	}
+
+	Joi.validate(options, schema, function (error) {
```
jagoda: It probably makes more sense to use the synchronous form of `Joi.validate()` here since this is a synchronous function. Check out the code examples in the [`validate()` documentation](https://github.com/hapijs/joi#validatevalue-schema-options-callback).

lib/services/awsAdapter.js
jagoda: Thoughts?

rrpod: I put this in a catch block and it logs incase of an error

test/unit/services/awsAdapter_spec.js
jagoda: Looks like it is still here to me.

rrpod: Interesting. Not still sure why the old code is visible. In the updated code https://github.com/rrpod/service-maker/blob/issue60_merge/test/unit/services/awsAdapter_spec.js (Line 177) it isn't there!

lib/services/awsAdapter.js
jagoda: As in you have more changes that haven't been pushed yet?

test/unit/services/awsAdapter_spec.js
jagoda: We should probably try to figure this out so that it doesn't end up getting merged. Unless there is a bug it GitHub, this line being visible here seems to indicate that it will get merged.

lib/services/awsAdapter.js
rrpod: Ok sorry I was looking at the error handling in the catch block. If updateInstance fails in the then block, it should ideally be caught in the catch block. I shall verify this.

PR #87, opened 2015-07-23
Implemented PUT as part of the REST API. Also fixed a regex issue with the schema for an instance.

arjunmadan:
To do: Add AWS integration to change the state of the instance on AWS

### DIFF
--- a/lib/models/Instance.js
+++ b/lib/models/Instance.js
@@ -6,11 +6,10 @@ var Joi     = require("joi");
 module.exports = Genesis.create("instance", {
 	index  : "id",
 	schema : Joi.object().keys({
-		id       : Joi.string().required(),
-		type     : Joi.string().optional().default("t2.micro"),
-		ami      : Joi.string().optional().default("ami-d05e75b8"),
-		state    : Joi.string().regex(/pending|ready|failed/).optional().default("pending"),
-		uri      : Joi.string().uri().optional().allow(null).default(null),
-		revision : Joi.number().integer().default(0)
+		id    : Joi.string().required(),
+		type  : Joi.string().optional().default("t2.micro"),
+		ami   : Joi.string().optional().default("ami-d05e75b8"),
+		state : Joi.string().regex(/\bfailed\b|\bready\b|\bpending\b/).optional().default("pending"),
+		uri   : Joi.string().uri().optional().allow(null).default(null),
 	})
 });
--- a/lib/models/Instance.js
+++ b/lib/models/Instance.js
@@ -9,7 +9,7 @@ module.exports = Genesis.create("instance", {
 		id    : Joi.string().required(),
 		type  : Joi.string().optional().default("t2.micro"),
 		ami   : Joi.string().optional().default("ami-d05e75b8"),
-		state : Joi.string().regex(/\bfailed\b|\bready\b|\bpending\b/).optional().default("pending"),
+		state : Joi.string().valid("failed", "ready", "pending").optional().default("pending"),
 		uri   : Joi.string().uri().optional().allow(null).default(null)
 	})
 });
--- a/lib/models/Instance.js
+++ b/lib/models/Instance.js
@@ -10,6 +10,6 @@ module.exports = Genesis.create("instance", {
 		type  : Joi.string().optional().default("t2.micro"),
 		ami   : Joi.string().optional().default("ami-d05e75b8"),
 		state : Joi.string().regex(/\bfailed\b|\bready\b|\bpending\b/).optional().default("pending"),
-		uri   : Joi.string().uri().optional().allow(null).default(null),
+		uri   : Joi.string().uri().optional().allow(null).default(null)
 	})
 });
--- a/lib/models/Instance.js
+++ b/lib/models/Instance.js
@@ -6,10 +6,11 @@ var Joi     = require("joi");
 module.exports = Genesis.create("instance", {
 	index  : "id",
 	schema : Joi.object().keys({
-		id    : Joi.string().required(),
-		type  : Joi.string().optional().default("t2.micro"),
-		ami   : Joi.string().optional().default("ami-d05e75b8"),
-		state : Joi.string().regex(/pending|ready|failed/).optional().default("pending"),
-		uri   : Joi.string().uri().optional().allow(null).default(null)
+		id       : Joi.string().required(),
+		type     : Joi.string().optional().default("t2.micro"),
+		ami      : Joi.string().optional().default("ami-d05e75b8"),
+		state    : Joi.string().regex(/pending|ready|failed/).optional().default("pending"),
+		uri      : Joi.string().uri().optional().allow(null).default(null),
+		revision : Joi.number().integer().default(0)
 	})
 });
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -169,7 +169,6 @@ exports.register = function (server, options, next) {
 
 			return instances.getInstance({ id : request.params.instanceId })
 			.then(function (instance) {
-				console.log(instance.revision);
 				var updatedInstance = new Instance({
 					id       : request.params.instanceId,
 					ami      : request.payload.ami,
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -161,6 +161,30 @@ exports.register = function (server, options, next) {
 			});
 		}
 	});
+
+	server.route({
+		method  : "PUT",
+		path    : "/v1/instances/{instanceId}",
+		handler : function (request, reply) {
+			var updatedInstance = new Instance({
+				id    : request.params.instanceId,
+				ami   : request.payload.ami,
+				type  : request.payload.type,
+				state : request.payload.state,
+				uri   : request.payload.uri
+			});
+
+			return instances.updateInstance(updatedInstance)
+			.then(function () {
+				reply("The instance was successfully updated.");
+			})
+			.catch (function (error) {
+				server.log([ "error", "rest" ], error.message);
+				reply(Boom.wrap(error));
+			});
+		}
+	});
+
 	next();
 };
 
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -199,6 +199,9 @@ exports.register = function (server, options, next) {
 						if (error.toString().match(/\bnot exist\b/)) {
 							reply(Boom.notFound("This instance does not exist."));
 						}
+						else if (error.toString().match(/\bconcurrency\b/)) {
+							reply(Boom.conflict("Conflict error. Ensure you have the latest document version"));
+						}
 						else {
 							reply(Boom.wrap(error));
 						}
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -164,33 +164,46 @@ exports.register = function (server, options, next) {
 
 	server.route({
 		method  : "PUT",
-		path    : "/v1/instances/{instanceId}",
+		path    : "/v1/instances/{id}",
 		handler : function (request, reply) {
-			instances.getInstance({ id : request.params.instanceId })
-			.then(function (response) {
 
-				var instance = new Instance({
-					id    : request.params.instanceId,
-					ami   : request.payload.ami,
-					type  : request.payload.type,
-					state : request.payload.state,
-					uri   : request.payload.uri
-				});
-				var Constructor = instance.constructor;
-				var updatedInstance = new Constructor(
-					_.merge(
-						_.clone(instance),
-						{ revision : response.revision }
-					)
-				);
-				return instances.updateInstance(updatedInstance);
-			})
-			.then(function (response) {
-				reply(response);
-			})
-			.catch (function (error) {
-				server.log([ "error", "rest" ], error.message);
-				reply(Boom.wrap(error));
+			var schema = Joi.object().keys({
+				type     : Joi.string(),
+				ami      : Joi.string(),
+				state    : Joi.string(),
+				uri      : Joi.string().allow(null),
+				revision : Joi.number().integer()
+			});
+
+			var payload = _.pick(request.payload, "ami", "type", "state", "uri", "revision");
+			Joi.validate(payload, schema, function (error) {
+				if (error) {
+					reply(Boom.badRequest("The parameters were entered incorrectly."));
+				}
+				else {
+					var instance = new Instance({
+						id       : request.params.id,
+						ami      : request.payload.ami,
+						type     : request.payload.type,
+						state    : request.payload.state,
+						uri      : request.payload.uri,
+						revision : request.payload.revision
+					});
+
+					return instances.updateInstance(instance)
+					.then(function (response) {
+						reply(response);
+					})
+					.catch (function (error) {
+						server.log([ "error", "rest" ], error.message);
+						if (error.toString().match(/\bnot exist\b/)) {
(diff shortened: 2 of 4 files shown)


lib/models/Instance.js
```diff
@@ -9,7 +9,7 @@ module.exports = Genesis.create("instance", {
 		id    : Joi.string().required(),
 		type  : Joi.string().optional().default("t2.micro"),
 		ami   : Joi.string().optional().default("ami-d05e75b8"),
-		state : Joi.string().regex(/pending|ready|failed/).optional().default("pending"),
+		state : Joi.string().regex(/\bfailed\b|\bready\b|\bpending\b/).optional().default("pending"),
```
jagoda: Does this imply that the `state` can have multiple values? Is there any reason that we wouldn't use Joi's [`allow()`](https://github.com/hapijs/joi#anyvalidvalue---aliases-only-equal) for this? Perhaps something like `Joi.string().allow("failed", "ready", "pending")`?

lib/plugins/rest.js
```diff
@@ -161,6 +161,40 @@ exports.register = function (server, options, next) {
...
+
+	server.route({
+		method  : "PUT",
+		path    : "/v1/instances/{instanceId}",
+		handler : function (request, reply) {
+			instances.getInstance({ id : request.params.instanceId })
+			.then(function (response) {
+
+				var instance = new Instance({
+					id    : request.params.instanceId,
```
jagoda: Just a thought, but if you renamed `{instanceId}` to `{id}` then you could simply do something like `var instance = new Instance(request.params);`.

lib/plugins/rest.js
```diff
@@ -161,6 +161,40 @@ exports.register = function (server, options, next) {
...
+					ami   : request.payload.ami,
+					type  : request.payload.type,
+					state : request.payload.state,
+					uri   : request.payload.uri
+				});
+				var Constructor = instance.constructor;
```
jagoda: This seems unnecessary since we explicitly used the `Instance` constructor in the previous expression. Could you explain why this was needed?

lib/plugins/rest.js
```diff
@@ -161,6 +161,40 @@ exports.register = function (server, options, next) {
...
+				var updatedInstance = new Constructor(
+					_.merge(
+						_.clone(instance),
+						{ revision : response.revision }
+					)
+				);
+				return instances.updateInstance(updatedInstance);
```
jagoda: This degrades the concurrency protection that `genesis` provides since it will make it look like all incoming requests are updating the current revision. Instead, the `revision` ID should be included on the incoming request payload. If the request payload includes the full `Instance` document, then we shouldn't need to call `getInstance()` either.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+			return server.registerAsync({
+				register : Rest,
+				options  : {
+					instances : instances
+				}
+			});
+		});
+
+		after(function () {
+			return server.stopAsync();
```
jagoda: The server was never started so we shouldn't need this.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+					.payload({
+						ami   : instance.ami,
+						type  : instance.type,
+						uri   : instance.uri,
+						state : "failed"
+					});
+					return request.inject(server);
+				})
+				.then(function (response) {
+					console.log(response);
```
jagoda: This dirties up the logs in the CI system. We should remove this before merging this code.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+			});
+		});
+
+		after(function () {
+			return server.stopAsync();
+		});
+
+		describe("setting the status of a created instance to failed", function () {
+
+			before(function () {
```
jagoda, line 673: I think it would more accurately reflect the real world flow here to use the `InstanceAdapter` only for setting up preconditions and to perform a `GET` request to fetch and update the instance. This should also simulate the `revision` behavior that we want.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+				})
+				.then(function (response) {
+					console.log(response);
+					result = response.result;
+				});
+			});
+
+			it("the status is set to failed", function () {
+				return instances.getInstance({ id : instanceID })
+				.then(function () {
```
jagoda: I don't understand why `getInstance()` is called here. It doesn't look like we use the result of this call. Rather, we are using the `result` value from the `before` block.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+					expect(result.id).to.equal(updatedInstance.id);
+					expect(result.ami).to.equal(updatedInstance.ami);
+					expect(result.type).to.equal(updatedInstance.type);
+					expect(result.state).to.equal(updatedInstance.state);
+					expect(result.uri).to.equal(updatedInstance.uri);
```
jagoda: Would it be simpler to use something like `expect(result).to.deep.equal(updatedInstance)`?

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+					});
+					return request.inject(server);
+				})
+				.then(function (response) {
+					console.log(response);
+					result = response.result;
+				});
+			});
+
+			it("the status is set to failed", function () {
```
jagoda, line 704: We also need to check the response code and that the instance state was updated in the database (can either do this directly via the `InstanceAdapter` or by performing a `GET` request).

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+						state : "failed"
+					});
+					return request.inject(server);
+				})
+				.then(function (response) {
+					result = response;
+				});
+			});
+
+			it("the status is set to failed", function () {
```
jagoda: This description seems inaccurate.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+				return instances.getInstance({ id : INVALID_QUERY })
+				.then(function () {
+					var error = JSON.parse(result.payload);
+					expect(error.statusCode).to.equal(500);
```
jagoda: The response here should use status code `404`.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+				})
+				.then(function (response) {
+					result = response;
+				});
+			});
+
+			it("the status is set to failed", function () {
+				return instances.getInstance({ id : INVALID_QUERY })
+				.then(function () {
+					var error = JSON.parse(result.payload);
```
jagoda: If you save the entire `response` object above, then you shouldn't need to parse the payload here. Instead, you can inspect `response.statusCode`. In fact, this is what we actually want to check since the HTTP status is not necessarily related to the payload.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+					expect(error.statusCode).to.equal(500);
+				});
+			});
+		});
```
jagoda, line 845: I think there is still an uncovered case for when the database fails during the update.

jagoda, line 845: I think there is still an uncovered case for when the update payload is malformed (does not validate).

jagoda, line 845: I think there is still an uncovered case for when two different clients try to update the instance simultaneously.

test/unit/plugins/rest_spec.js
jagoda: In fact, we should have an assertion in this test case that ensures the opposite is true. If the URL is incorrect, the underlying status should not be modified.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,97 @@ describe("The Rest plugin", function () {
...
+		after(function () {
+			return server.stopAsync();
+		});
+
+		describe("setting the status of a created instance to failed", function () {
+
+			before(function () {
+				return instances.createInstance()
+				.then(function (instance) {
+					instanceID = instance.id;
```
jagoda, line 677: I don't think that this is necessary.

test/unit/plugins/rest_spec.js
arjunmadan, line 845: For the simultaneous update case the expected behavior should be the first request succeeding and the second failing, right?

jagoda, line 845: Ordering is probably not well defined, but you have the right idea. One request should succeed and one should fail.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					return instances.getInstance({ id : instance.id });
+				})
+				.then(function (instance) {
+					updatedInstance = new Instance({
+						id       : instance.id,
+						ami      : instance.ami,
+						type     : instance.type,
+						state    : "failed",
+						uri      : instance.uri,
+						revision : 1
```
jagoda: Is this needed? I would expect the normal flow to get the revision from the previous `.getInstance()` call.

jagoda: Ah, I see. This is primarily for comparison. It still probably makes the most sense to do something like `instance.revision + 1` here.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					});
+					var request = new Request("PUT", "/v1/instances/" + instance.id).mime("application/json")
+					.payload({
+						ami      : instance.ami,
+						type     : instance.type,
+						uri      : instance.uri,
+						state    : "failed",
+						revision : 0
```
jagoda: Similar to the previous comment, I would expect this to come from the fetched instance rather than being hard-coded.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					revision : 0
+				});
+				return request.inject(server)
+				.then(function (response) {
+					result = response;
+				});
+			});
+
+			it("a 400 error is thrown", function () {
+				expect(result.statusCode).to.equal(400);
```
jagoda, line 761: I'm surprised that this works this way. I would have expected the invalid URL to result in a `404` response prior to attempting to validate the payload. Would it make more sense to at least have the URL be a valid URL in this test case? That way the intent of the test case is very clear and well-defined.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					return instances.getInstance({ id : instance.id });
+				})
+				.then(function (instance) {
+					var firstRequest = new Request("PUT", "/v1/instances/" + instance.id).mime("application/json")
+					.payload({
+						ami      : instance.ami,
+						type     : instance.type,
+						uri      : null,
+						state    : "ready",
+						revision : 0
```
jagoda: Similar to above, would it make more sense to pull this off of the `instance` object?

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					});
+
+					var secondRequest = new Request("PUT", "/v1/instances/" + instance.id).mime("application/json")
+					.payload({
+						ami      : instance.id,
+						type     : instance.type,
+						uri      : null,
+						state    : "failed",
+						revision : 0
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/87/files#r35551684).

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+
+					firstRequest.inject(server)
+					.then(function (response) {
+						firstResponse = response;
+					});
+
+					secondRequest.inject(server)
+					.then(function (response) {
+						secondResponse = response;
+					});
```
jagoda: How do we know that both requests have completed before we perform the assertions?

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+				});
+
+			});
+
+			it("one succeeds while the other fails", function () {
+				if (firstResponse.statusCode === 200 && secondResponse.statusCode === 500) {
```
jagoda: In the case of an error, the status code should be `409`. See the [HTTP spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.10) for more details.

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+					instances.getInstance({ id : instanceID })
+					.then(function (instance) {
+						expect(instance.state).to.equal("ready");
+					});
+				}
+				else if (secondResponse.statusCode === 200 && firstResponse.statusCode === 500) {
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/87/files#r35551951).

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+				return instances.getInstance({ id : instance.id });
+			})
+			.then(function (instance) {
+				updatedInstance = new Instance({
+					id       : instance.id,
+					ami      : instance.ami,
+					type     : instance.type,
+					state    : "failed",
+					uri      : instance.uri,
+					revision : 1
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/87/files#r35551156).

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +647,233 @@ describe("The Rest plugin", function () {
...
+				});
+				var request = new Request("PUT", "/v1/instances/" + instance.id).mime("application/json")
+				.payload({
+					ami      : instance.ami,
+					type     : instance.type,
+					uri      : instance.uri,
+					state    : "failed",
+					revision : 0
```
jagoda: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/87/files#r35551211).

test/unit/plugins/rest_spec.js
```diff
@@ -646,4 +648,210 @@ describe("The Rest plugin", function () {
...
+				});
+
+			});
+
+			it("one succeeds while the other fails", function () {
+				var failed = _.some(responses, function (responses) {
+					return responses.statusCode === 409;
+				});
+
+				expect (failed, "concurrency").to.be.true;
```
jagoda, line 797: This is nice! :+1:

jagoda:
Looks good to me. Thanks for all the work on this!

PR #88, opened 2015-07-28
Completed terminating an instance with PUT.

arjunmadan:
- instance state is set to terminated once it is terminated on the EC2 console.
- instance state is updated to failed in case AWS fails.

### DIFF
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -197,57 +197,47 @@ exports.register = function (server, options, next) {
 						uri      : request.payload.uri,
 						revision : request.payload.revision
 					});
-
-					return instances.updateInstance(instance)
-					.then(function (response) {
-						switch (response.state) {
-							case ("terminating") : {
+					switch (instance.state) {
+						case ("terminating") : {
+							return instances.updateInstance(instance)
+							.then(function (response) {
 								reply(response);
-								return awsAdapter.terminateInstances(response.id)
-								.then(function () {
-									return instances.getInstance({ id : response.id });
-								})
-								.then(function (instance) {
-									instance = new Instance({
-										id       : instance.id,
-										ami      : instance.ami,
-										type     : instance.type,
-										state    : "terminated",
-										uri      : instance.uri,
-										revision : instance.revision
-									});
-									return instances.updateInstance(instance);
-								})
-								.then(function (response) {
-									console.log(response);
+								return awsAdapter.terminateInstances(response.id);
+							})
+							.then(function () {
+								return instances.getInstance({ id : request.params.id });
+							})
+							.then(function (response) {
+								instance = new Instance({
+									id       : response.id,
+									ami      : response.ami,
+									type     : response.type,
+									state    : "terminated",
+									uri      : response.uri,
+									revision : response.revision
 								});
-							}
-							case ("starting") : {
-								//To do: Add startInstance()
-								break;
-							}
-							case ("stopping") : {
-								//To do: Add stopInstance()
-								break;
-							}
-							default : {
-								server.log([ "error", "rest" ], "Incorrect parameters");
-								break;
-							}
+								return instances.updateInstance(instance);
+							})
+							.catch (function (error) {
+								server.log([ "error", "rest" ], error.message);
+								if (error.toString().match(/\bnot exist\b/)) {
+									reply(Boom.notFound("This instance does not exist."));
+								}
+								else if (error.toString().match(/\bconcurrency\b/)) {
+									reply(Boom.conflict("Conflict error. Ensure you have the latest document version"));
+								}
+								else {
+									reply(Boom.wrap(error));
+								}
+							});
 						}
-					})
-					.catch (function (error) {
-						server.log([ "error", "rest" ], error.message);
-						if (error.toString().match(/\bnot exist\b/)) {
-							reply(Boom.notFound("This instance does not exist."));
+
+						default : {
+							server.log([ "error", "rest" ], "Incorrect parameters");
+							reply(Boom.badRequest("The state entered is not valid."));
+							break;
 						}
-						else if (error.toString().match(/\bconcurrency\b/)) {
-							reply(Boom.conflict("Conflict error. Ensure you have the latest document version"));
-						}
-						else {
-							reply(Boom.wrap(error));
-						}
-					});
+					}
 				}
 			});
 		}
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -200,6 +200,21 @@ exports.register = function (server, options, next) {
 
 					return instances.updateInstance(instance)
 					.then(function (response) {
+						switch (response.state) {
+							case ("failed") : {
+								return awsAdapter.terminateInstances(response.id)
+								.then(function (response) {
+									console.log(response);
+								});
+								break;
+							}
+							case ("terminating") : {
+								break;
+							}
+							default : {
+								break;
+							}
+						}
 						reply(response);
 					})
 					.catch (function (error) {
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -201,17 +201,15 @@ exports.register = function (server, options, next) {
 					return instances.updateInstance(instance)
 					.then(function (response) {
 						switch (response.state) {
-							case ("failed") : {
+							case ("terminating") : {
 								return awsAdapter.terminateInstances(response.id)
 								.then(function (response) {
 									console.log(response);
 								});
 								break;
 							}
-							case ("terminating") : {
-								break;
-							}
 							default : {
+								server.log([ "", "rest" ], "Incorrect parameters");
 								break;
 							}
 						}
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -202,12 +202,25 @@ exports.register = function (server, options, next) {
 					.then(function (response) {
 						switch (response.state) {
 							case ("terminating") : {
+								reply(response);
 								return awsAdapter.terminateInstances(response.id)
+								.then(function () {
+									return instances.getInstance({ id : response.id });
+								})
+								.then(function (instance) {
+									instance = new Instance({
+										id       : instance.id,
+										ami      : instance.ami,
+										type     : instance.type,
+										state    : "terminated",
+										uri      : instance.uri,
+										revision : instance.revision
+									});
+									return instances.updateInstance(instance);
+								})
 								.then(function (response) {
-									//Handle background process
 									console.log(response);
 								});
-								break;
 							}
 							case ("starting") : {
 								//To do: Add startInstance()
@@ -218,11 +231,10 @@ exports.register = function (server, options, next) {
 								break;
 							}
 							default : {
-								server.log([ "", "rest" ], "Incorrect parameters");
+								server.log([ "error", "rest" ], "Incorrect parameters");
 								break;
 							}
 						}
-						reply(response);
 					})
 					.catch (function (error) {
 						server.log([ "error", "rest" ], error.message);
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -202,47 +202,52 @@ exports.register = function (server, options, next) {
 							return instances.updateInstance(instance)
 							.then(function (response) {
 								reply(response);
-								return awsAdapter.terminateInstances(response.id);
-							})
-							.then(function () {
-								return instances.getInstance({ id : request.params.id });
-							})
-							.then(function (response) {
-								instance = new Instance({
-									id       : response.id,
-									ami      : response.ami,
-									type     : response.type,
-									state    : "terminated",
-									uri      : null,
-									revision : response.revision
+								awsAdapter.terminateInstances(request.params.id)
+								.then(function () {
+									return instances.getInstance({ id : request.params.id });
+								})
+								.then(function (response) {
+									instance = new Instance({
+										id       : response.id,
+										ami      : response.ami,
+										type     : response.type,
+										state    : "terminated",
+										uri      : null,
+										revision : response.revision
+									});
+									return instances.updateInstance(instance);
+								})
+								.catch (function (error) {
+									if (error.toString().match(/\bdocument\b|\bconcurrency\b|\bdatabase\b/)) {
+										server.log([ "error", "rest", "put" ], error.message);
+									}
+									else {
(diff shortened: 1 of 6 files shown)


lib/plugins/rest.js
```diff
@@ -189,23 +189,62 @@ exports.register = function (server, options, next) {
...
+									id       : response.id,
+									ami      : response.ami,
+									type     : response.type,
+									state    : "terminated",
+									uri      : null,
+									revision : response.revision
+								});
+								return instances.updateInstance(instance);
+							})
+							.catch (function (error) {
```
jagoda, line 238: This error handling probably needs to be reworked. I believe there are cases where we could end up calling `reply()` twice.

lib/services/awsAdapter.js
```diff
@@ -44,6 +44,35 @@ function AWSAdapter (ec2) {
 			return result;
 		});
 	});
+
+	this.terminateInstances = Bluebird.method(function (instanceId) {
```
jagoda: Is this `Bluebird.method()` call necessary? It looks like everything in this method is already in a promise chain.

test/unit/plugins/rest_spec.js
```diff
@@ -678,20 +682,31 @@ describe("The Rest plugin", function () {
...
+
 					updatedInstance = new Instance({
 						id       : instance.id,
 						ami      : instance.ami,
 						type     : instance.type,
-						state    : "failed",
+						state    : "terminated",
 						uri      : instance.uri,
-						revision : instance.revision + 1
+						revision : instance.revision + 2
```
jagoda, line 702: I'm curious why we check for `instance.revision + 2`?

test/unit/plugins/rest_spec.js
```diff
@@ -765,10 +784,21 @@ describe("The Rest plugin", function () {
...
 				return instances.createInstance()
 				.then(function (instance) {
+					terminateInstancesStub = Sinon.stub(awsAdapter, "terminateInstances").returns(
+						Bluebird.resolve({
+							id       : instance.id,
+							ami      : instance.ami,
+							type     : instance.type,
+							state    : "terminated",
+							uri      : instance.uri,
+							revision : instance.revision + 2
```
jagoda, line 805: See [earlier comment](https://github.com/inetCatapult/service-maker/pull/88/files#r35686727).

test/unit/plugins/rest_spec.js
```diff
@@ -797,18 +831,102 @@ describe("The Rest plugin", function () {
...
+			var instanceID;
+			var responseCode;
+			var originalInstance;
+
+			before(function () {
+
+				return instances.createInstance()
+				.then(function (instance) {
+					instanceID = instance.id;
+					return instances.getInstance({ id : instance.id });
```
jagoda: What is the rationale for calling `instances.getInstance()` here? Does `instance.createInstance()` not return what you need?

test/unit/plugins/rest_spec.js
```diff
@@ -797,18 +831,102 @@ describe("The Rest plugin", function () {
...
+				})
+				.then(function (instance) {
+					originalInstance = {
```
jagoda: Would `_.clone()` be simpler here?

jagoda: Do we even need to clone the instance at all? Can we simply copy the reference?

test/unit/plugins/rest_spec.js
arjunmadan, line 702: It shouldn't really matter since I'm stubbing `terminateInstances()`, but the reason I've done this is because when this is not stubbed the `state` is first set to `terminating` and when AWS says it's been successfully terminated it is set to `terminated`. Since `updateInstance()` is called twice the final revision would be the original + 2.

test/unit/plugins/rest_spec.js
```diff
@@ -797,18 +831,102 @@ describe("The Rest plugin", function () {
...
+					});
+					return request.inject(server);
+				})
+				.then(function (response) {
+					responseCode = response.statusCode;
+				});
+			});
+
+			it("throws an error", function () {
+				return instances.getInstance({ id : instanceID })
```
jagoda, line 874: I don't understand why `instances.getInstance()` is called here. Can you please explain?

jagoda, line 874: Is it to verify that the instance is not modified in the event of an error? If that is the case, it would probably be clearer (and more self-descriptive) to separate this out into it's own assertion (`it` block).

test/unit/plugins/rest_spec.js
```diff
@@ -797,18 +831,102 @@ describe("The Rest plugin", function () {
...
+		});
+
+		describe("when there is an AWS error", function () {
+
+			var terminateInstancesStub;
+			var instanceID;
+
+			before(function () {
+				terminateInstancesStub = Sinon.stub(awsAdapter, "terminateInstances").rejects(new Error("AWS Error"));
+				instances.createInstance()
```
jagoda: I think there is a potential timing problem here.

test/unit/plugins/rest_spec.js
```diff
@@ -797,18 +831,102 @@ describe("The Rest plugin", function () {
...
+					});
+					return request.inject(server);
+				});
+			});
+
+			after(function () {
+				terminateInstancesStub.restore();
+			});
+
+			it("returns the instance with state set to failed", function () {
```
jagoda, line 917: Since this is the test suite for the REST API we should probably perform some assertions on the HTTP response and not just the DB model.

test/unit/services/awsAdapter_spec.js
```diff
@@ -5,7 +5,9 @@ var expect     = require("chai").expect;
 var Bluebird   = require("bluebird");
 var Sinon      = require("sinon");
 var AWS        = require("aws-sdk");
-var ec2        = new AWS.EC2();
+var ec2        = new AWS.EC2({
```
jagoda: This should probably happen inside a `describe` block since it won't get garbage collected otherwise.

test/unit/services/awsAdapter_spec.js
```diff
@@ -5,7 +5,9 @@ var expect     = require("chai").expect;
...
+		region : "us-east-1"
```
jagoda: Can you explain the rationale for hard-coding a region here? I wouldn't expect this to be necessary (or desirable).

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +152,146 @@ describe("The AwsAdapter class ", function () {
...
+					var data = { Reservations : [ { Instances : [ {
+						InstanceId : VALID_AWS_ID,
+						state      : "terminated"
+					} ] } ] };
+					return Bluebird.resolve(data);
+				});
+
+				return awsAdapter.terminateInstances(VALID_ID)
+				.then(function (response) {
+					result = response;
```
jagoda, line 492: I don't think that this ever gets used in the assertions below. Do we care about this?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +152,146 @@ describe("The AwsAdapter class ", function () {
...
+				expect(waitForStub.args[ 0 ][ 1 ].Filters[ 0 ].Name).to.equal("tag:ID");
+				expect(waitForStub.args[ 0 ][ 1 ].Filters[ 0 ].Values[ 0 ]).to.equal(VALID_ID);
+
+				expect(result.Reservations[ 0 ].Instances[ 0 ].InstanceId).to.equal(VALID_AWS_ID);
+				expect(result.Reservations[ 0 ].Instances[ 0 ].state).to.equal("terminated");
+			});
+
+		});
+
+		describe("when the instance has already been terminated (doesn't exist)", function () {
```
jagoda, line 517: I want to make sure that the intention of this test case is clear. This message (and the error message below) seems to imply that this test case should cover the case where the instance does not exist. However, in that case I would expect the `describeInstance()` call to fail since there is no instance to describe. Perhaps we should have two different test cases for the different scenarios?

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,4 +152,146 @@ describe("The AwsAdapter class ", function () {
...
+
+			after(function () {
+				describeInstancesStub.restore();
+				terminateInstancesStub.restore();
+			});
+
+			it("throws an error", function () {
+				expect(describeInstancesStub.args[ 0 ][ 0 ].Filters[ 0 ].Name).to.equal("tag:ID");
+				expect(describeInstancesStub.args[ 0 ][ 0 ].Filters[ 0 ].Values[ 0 ]).to.equal(VALID_ID);
+				expect(result.message).to.contain("not exist");
```
jagoda: We should probably verify that `terminateInstance()` was called with the correct arguments since it's stub is supplying the error value here.

test/unit/services/awsAdapter_spec.js
arjunmadan, line 517: From what I understand I think that `describeInstances()` returns an empty `Instances` array if it can't find the instance. That's what I've tried to replicate here. I'll just verify that this is actually the case.

test/unit/plugins/rest_spec.js
jagoda, line 702: If it doesn't really matter than would it be more correct to check for `response.revision >= instance.revision + 1`?

test/unit/services/awsAdapter_spec.js
arjunmadan, line 492: I think I'm using it to verify that the InstanceId returned is the same one passed in and the instance state is set to terminated.

test/unit/services/awsAdapter_spec.js
jagoda, line 517: Would be good to verify. But also, your stub implementation below seems to return a populated array. I think this will be confusing to others if it is also intended to cover the case where AWS returns an empty array.

test/unit/services/awsAdapter_spec.js
jagoda, line 492: Oops! You are correct -- I missed it. Sorry about that!

test/unit/services/awsAdapter_spec.js
arjunmadan, line 517: I just noticed that and am now returning an empty array. This will be reflected once I update the PR.

jagoda, line 517: Awesome.

lib/plugins/rest.js
arjunmadan, line 238: I think the way I have it worked out right now is that the error cases only reply when `updateInstance()` fails the first time. The rest of the work happens in the background and that's what the `else` block is for. I'll verify this to make sure this is actually the case.

jagoda, line 238: The problem is that the error handling is happening in the same promise chain that the `reply()` call happens in. It may be an obscure edge case, but consider what would happen if the `reply()` call on line 196 succeeds but then the database goes down. In this case the code will attempt to call `reply()` again on line 222. I don't think that hapi will allow this and it may very well obscure what the actual problem is.

arjunmadan, line 238: Would it also make sense to nest the promise chain in this case? Once the reply is sent out I can create a nested promise chain that would just log errors if any, and not send a reply.

arjunmadan:
I think I messed up some of the history while doing the rebase but the code appears to be the latest version. 
Edit: Everything seems fine now.

test/unit/plugins/rest_spec.js
```diff
@@ -821,20 +948,80 @@ describe("The Rest plugin", function () {
...
+					Bluebird.resolve({
+						id       : instance.id,
+						ami      : instance.ami,
+						type     : instance.type,
+						state    : "terminating",
+						uri      : instance.uri,
+						revision : instance.revision + 1
+					})
+				);
+				updateStub.onCall(1).rejects(new Error("Connection to the database fails."));
```
jagoda, line 996: We still need some way to verify that this is executed before we run the assertion. Otherwise, it may be that the instance is still in the "terminating" state simply because the async update hasn't actually happened yet.

test/unit/plugins/rest_spec.js
```diff
@@ -845,13 +1032,63 @@ describe("The Rest plugin", function () {
...
+			.then(function (response) {
+				result = response.result;
+			});
+		});
+
+		after(function () {
+			terminateInstancesStub.restore();
+			getStub.restore();
+		});
+
```
jagoda, line 1089: What is the expected response from the API in this case?

arjunmadan, line 1089: This is the case where it can update successfully the first time setting the `state` to `terminating`. `terminateInstance()` and the database fail after this so the `state` at the end should still be `terminating`

test/unit/services/awsAdapter_spec.js
```diff
@@ -455,4 +457,136 @@ describe("The AwsAdapter class ", function () {
...
+		});
+
+		describe("when the instance has already been terminated (doesn't exist)", function () {
+
+			var result;
+			var awsAdapter = new AwsAdapter(awsOptions);
+
+			before(function () {
+				describeInstancesStub = Sinon.stub(ec2, "describeInstancesAsync")
+				.rejects(new Error("InvalidInstance.NotFound"));
```
jagoda, line 524: This confuses me because I thought we had a discussion the other day about how in this case the `describeInstances()` method returns an empty result set and not an error. Was that incorrect?

arjunmadan, line 524: Yup, I just verified that. I'd forgotten to change it before.

test/unit/plugins/rest_spec.js
jagoda, line 1089: So should we assert that the status code is 200?

test/unit/services/awsAdapter_spec.js
jagoda, line 524: Is the same true when the "instance has already been terminated"? Is there any case where we can get the empty result set?

arjunmadan, line 524: The API docs mention that the array returned has size 1 or more.

jagoda, line 524: I see. I'd like to re-trace how/why we originally set this to an empty array. I feel like I asked about this section of the code in a previous review and was told that `describeInstances()` could return and empty array. Good topic for the scrum.

jagoda:
I think this is good to go now. Thanks for all the work on this!

PR #90, opened 2015-07-30
Heroku deployment.

jagoda:
This change enables continuous delivery to Heroku. The default URL
is http://service-maker-demo.herokuapp.com.


jagoda:
Thanks! Looks like the service is live at http://service-maker-demo.herokuapp.com and is successfully deploying on merges to `v1.0` :-)

PR #91, opened 2015-07-31
Instances now get assigned a default security group - "service-maker"

arjunmadan:
- if the group doesn't exist on the AWS console, it is created and a rule for SSH is added.
- if the group exists, nothing is done with it.

### DIFF
--- a/test/unit/services/awsAdapter_spec.js
+++ b/test/unit/services/awsAdapter_spec.js
@@ -106,15 +106,18 @@ describe("The AwsAdapter class ", function () {
 				authorizeSecurityGroupIngressStub.restore();
 			});
 
-			it("creates the security group and returns a new instance with the ami and type provided", function () {
-
+			it("creates the security group", function () {
 				expect(createSecurityGroupStub.args[ 0 ][ 0 ].GroupName).to.equal(DEFAULT_SG_NAME);
+			});
 
+			it("adds rules for SSH to the security group", function () {
 				expect(authorizeSecurityGroupIngressStub.args[ 0 ][ 0 ].GroupId).to.equal(DEFAULT_GROUP_ID);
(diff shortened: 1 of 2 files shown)


test/unit/services/awsAdapter_spec.js
```diff
@@ -65,8 +74,77 @@ describe("The AwsAdapter class ", function () {
...
+			});
+
+			after(function () {
+				runInstancesStub.restore();
+				beginPollingStub.restore();
+				createSecurityGroupStub.restore();
+				authorizeSecurityGroupIngressStub.restore();
+			});
+
+			it("creates the security group and returns a new instance with the ami and type provided", function () {
```
a-type: It would be more helpful when reviewing tests to separate the validation of stub parameters into different `it` blocks. For instance, if the wrong parameters were passed to both the first and second stub, this `it` block will fail, but only print output for the first failure. After fixing the first stub, the block would fail again for the second. By separating `it` blocks to be specifically testing each stub, we can see all of the failures on the first pass and fix them more efficiently, without any hidden failures.

a-type:
Looks pretty good, I have one suggestion for the test block structure.

test/unit/services/awsAdapter_spec.js
```diff
@@ -84,21 +162,84 @@ describe("The AwsAdapter class ", function () {
 			after(function () {
 				runInstancesStub.restore();
 				beginPollingStub.restore();
+				createSecurityGroupStub.restore();
+				authorizeSecurityGroupIngressStub.restore();
 			});
 
 			it("returns a new instance with the ami and type provided", function () {
```
a-type: See [above](https://github.com/inetCatapult/service-maker/pull/91/files#r36096653).

a-type:
Looks better, thanks!

PR #92, opened 2015-08-03
Stopping an instance with PUT

arjunmadan:
- stop an instance on EC2 if it is running.
- if it has already been stopped, do nothing.

### DIFF
--- a/test/unit/plugins/rest_spec.js
+++ b/test/unit/plugins/rest_spec.js
@@ -1494,7 +1494,7 @@ describe("The Rest plugin", function () {
 
 	});
 
-	describe("creating awsAdapter object", function () {
+	describe("creating awsAdapter object when the options passed are invalid", function () {
 		var server     = new Hapi.Server();
 		var awsAdapter = {};
 		var result;
--- a/test/unit/plugins/rest_spec.js
+++ b/test/unit/plugins/rest_spec.js
@@ -667,7 +667,7 @@ describe("The Rest plugin", function () {
 			});
 		});
 
-		describe("setting the status of a created instance to terminated", function () {
+		describe("setting the state of a created instance to terminated", function () {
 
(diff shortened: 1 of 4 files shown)


arjunmadan:
Sorry about the size of this PR. Most of the changes in rest_spec.js are just with the how the test cases were structured. I think they're organized much better now.

test/unit/plugins/rest_spec.js
```diff
@@ -675,425 +675,853 @@ describe("The Rest plugin", function () {
...
+						result = response.result;
+					});
+				});
+
+				after(function () {
+					terminateInstancesStub.restore();
+					updateStub.restore();
+				});
+
+				it("leaves the state unchanged(terminating)", function () {
```
a-type, line 1025: Is this the correct test result? The way I read the code, if there is an error after requesting a terminated instance, the instance should attempt to update to the "failed" state. I know that, due to the way you're using stubs and `done`, you don't actually reach that update. But, I think it may be possible to if you make some adjustments to the strategy you've developed.

a-type:
I think, after a second look, there's more depth that could be added to your test cases surrounding the behavior which happens after that reply is sent off. You should be able to use the same strategy you came up with, with some modifications, to be able to test more of that behavior before finishing the test case.

test/unit/plugins/rest_spec.js
arjunmadan, line 1025: This was for the case where the instance was set to `terminating` and the database fails after this - preventing it from being updated to `failed`.

a-type:
Nice job adapting the solution. There's just one unnecessary bit I think we could take out. Sometimes it's as much about what you don't need as what you do.

a-type:
Thanks for your hard work on this. I'm glad we could get the testing to be pretty comprehensive.

PR #94, opened 2015-08-04
Delete Rest API

rrpod:


a-type:
This PR will need to be rebased on top of @trueblue2704 's changes after they are accepted. I'll go ahead and note, though, that it will need to be updated with similar in-depth test cases which test the behavior which handles failures during the delete process.

a-type:
Ok, looks like we're ready to rebase and resolve.

PR #95, opened 2015-08-05
Starting an instance with a PUT request.

arjunmadan:
start a previously stopped instance and update the public IP address of the instance.


lib/services/awsAdapter.js
```diff
@@ -225,26 +225,48 @@ function AWSAdapter (options) {
...
-				} ]
-			};
 			return ec2.waitForAsync("instanceStopped", params);
 		})
 		.then(function (response) {
 			return response;
 		});
 	};
+
+	this.startInstances = function (instanceId) {
```
a-type: When the instance is started again, should you begin a new SSH polling task? Do we know whether the instance is immediately available, or if there is a delay in being able to SSH?

arjunmadan: That makes sense. I'll add it.

I'm going to wait for Rohit to get his fix for the polling approved before adding it. Right now I think it will fail because while updating it will lead to concurrency issues.

arjunmadan:
Now polling after the instance is started. This should only be merged once #98 is merged as it depends on the concurrency fix from that PR.

a-type:
Ok, seems good. Thanks.

PR #96, opened 2015-08-05
Delete Rest API

rrpod:


test/unit/plugins/rest_spec.js
```diff
@@ -2021,4 +2021,411 @@ describe("The Rest plugin", function () {
...
+			after(function () {
+				terminateInstancesStub.restore();
+			});
+
+			it("returns the instance state with state set to terminating", function () {
+				expect(result.state).to.equal("terminating");
+				expect(responseCode).to.equal(200);
+			});
+
+			it("returns the instance with state set to failed", function () {
```
a-type: These two `it` statements are a bit conflicting, and could be cleared up. I noticed there's the same problem on line 1534 as well, if you wouldn't mind fixing them both. The second `it` should be talking about the database model, not anything returned.

lib/plugins/rest.js
```diff
@@ -310,6 +264,86 @@ exports.register = function (server, options, next) {
...
+				return terminateInstance(updatedInstance, reply);
+			})
+			.catch(function (error) {
+				server.log([ "error", "rest", "NotFound" ], error);
+				reply(Boom.wrap(error));
+			});
+		}
+	});
+
+	function terminateInstance(instance, reply) {
```
a-type, line 296: Good thought to encapsulate this into a function.

PR #97, opened 2015-08-06
Use a custom security group

arjunmadan:
- creates the group and adds SSH if the group doesn't exist.
- uses an existing group if it does exist.
- falls back to the default group - "service-maker" if no group is specified.

### DIFF
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -27,6 +27,11 @@ exports.register = function (server, options, next) {
 		throw error;
 	}
 
+	if (process.env.AWS_DEFAULT_SECURITY_GROUP === undefined) {
+		server.log([ "error", "rest" ], "The default security group is not set");
+		throw new Error("Default Security Group not set");
+	}
+
 	server.route({
 		method  : "GET",
 		path    : "/",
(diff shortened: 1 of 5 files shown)


lib/plugins/rest.js
```diff
@@ -27,6 +27,11 @@ exports.register = function (server, options, next) {
 		throw error;
 	}
 
+	if (process.env.AWS_DEFAULT_SECURITY_GROUP === undefined) {
```
a-type: I don't think this needs to be a required parameter. We can provide a hard-coded default, and optionally allow the user to customize the default via the variable.

test/unit/plugins/rest_spec.js
```diff
@@ -36,15 +38,18 @@ describe("The Rest plugin", function () {
...
 		before(function () {
 			server.connection();
+			environment.set("AWS_DEFAULT_SECURITY_GROUP", VALID_SEC_GROUP);
+
 			return server.registerAsync(Rest);
 		});
 
 		after(function () {
-			return server.stopAsync();
+			environment.restore();
```
a-type: We lost the server stop here, it looks like.

test/unit/plugins/rest_spec.js
```diff
@@ -151,103 +225,171 @@ describe("The Rest plugin", function () {
...
```
a-type: Lost server stop here as well. Is there a reason for that?

test/unit/plugins/rest_spec.js
```diff
@@ -258,8 +400,8 @@ describe("The Rest plugin", function () {
 		});
 
 		after(function () {
+			environment.restore();
```
a-type, line 431: And here. There are more, but I think you get the idea.

test/unit/plugins/rest_spec.js
arjunmadan: I realized we were never actually starting the server.

a-type:
Ok, looks good.

PR #98, opened 2015-08-06
Added Cases to handle concurrency during sshpolling

rrpod:


test/unit/services/awsAdapter_spec.js
```diff
@@ -564,6 +565,157 @@ describe("The AwsAdapter class ", function () {
...
+				});
+			});
+
+			after(function () {
+				SshPollingStub.restore();
+				getPublicIPAddressStub.restore();
+				getInstanceStub.restore();
+			});
+
+			it("fails to update as it has wrong revision number", function () {
```
a-type: I'm not sure I understand this test case. If the `getInstance` indicates that the instance has not been updated, why does the operation fail?

rrpod: This is a test case where a getInstance is performed, an update instance is done (by a PUT request) and then an updateInstance is done based on the details received from the getInstance.

a-type: Perhaps that could be clarified better in the description. Just based on the description alone, it looks like it should succeed.

rrpod: I wanted to make it more descriptive. However multi-line descriptions aren't allowed with the current settings of jslint! Shall I write it in 2 describe blocks?

test/unit/services/awsAdapter_spec.js
```diff
@@ -564,6 +565,157 @@ describe("The AwsAdapter class ", function () {
...
+					SshPollingStub = Sinon.stub(sshAdapter,"SshPolling", function () {
+						return Bluebird.reject("Simulated Failure.");
+					});
+
+					getPublicIPAddressStub = Sinon.stub(awsAdapter,"getPublicIPAddress", function () {
+						return Bluebird.resolve(VALID_IP_ADDRESS);
+					});
+
+					getInstanceStub = Sinon.stub(instances, "getInstance", function () {
+						return instances.updateInstance(instance)
```
a-type: Would be nice to have some inline documentation here to explain what is happening.

test/unit/services/awsAdapter_spec.js
a-type: You could break out `"faces an error while polling"` into its own block and make the others sub-descriptions of that. I wouldn't focus so much on `getInstance` in the description, and focus more on the fact that an out-of-band update has occurred.

a-type:
Thanks!

PR #99, opened 2015-08-10
Added test cases for begin Polling function

rrpod:
Now validating the parameter passed to beginPolling function.

### DIFF
--- a/lib/services/awsAdapter.js
+++ b/lib/services/awsAdapter.js
@@ -46,11 +46,12 @@ function AWSAdapter (options) {
 		var groupName = process.env.AWS_DEFAULT_SECURITY_GROUP || "service-maker";
 
 		var result = {
-			id    : instance.id,
-			ami   : instance.ami,
-			type  : instance.type,
-			state : instance.state,
-			uri   : null,
+			id       : instance.id,
+			ami      : instance.ami,
+			type     : instance.type,
+			state    : instance.state,
+			uri      : null,
+			revision : instance.revision
 		};
 
 		var securityGroupParams = {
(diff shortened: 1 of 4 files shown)


lib/services/awsAdapter.js
```diff
@@ -118,6 +119,24 @@ function AWSAdapter (options) {
 	this.beginPolling = function (result) {
 		var self = this;
 		var instanceIPAddress;
+
+		var objectSchema = Joi.object().keys({
```
a-type: I think this is a bit of duplication. Instead of creating a new schema to test for validity, we could just instantiate a new `Instance`, which will do that for us. We can then pass the `Instance`'s id to the `SshPolling` function.

lib/services/awsAdapter.js
```diff
@@ -46,11 +46,12 @@ function AWSAdapter (options) {
 		var groupName = process.env.AWS_DEFAULT_SECURITY_GROUP || "service-maker";
 
 		var result = {
```
a-type: As discussed previously, `_.clone` will save you some maintenance in the future by copying all properties for you.

test/unit/services/awsAdapter_spec.js
```diff
@@ -622,13 +640,15 @@ describe("The AwsAdapter class ", function () {
...
 				var getInstanceStub;
-				var instanceProp;
+				var instanceProp = { };
 				var result;
 
 				before(function () {
 					return instances.createInstance()
 					.then(function (instance) {
-						instanceProp = instance;
+						instanceProp = JSON.parse(JSON.stringify(instance));
```
a-type: `_.clone` is also more performant for these kinds of operations.

a-type:
Looks good, thanks.

PR #100, opened 2015-08-10
Documentation

arjunmadan:
Updated API.md to reflect the current state of the API.
Added use cases for GET, PUT and DELETE to STORIES.md


a-type:
Wow, that's a lot of user stories added! Stories like that will be useful guides to creating integration tests if we move in that direction. Thanks for this.

PR #102, opened 2015-08-12
Use a custom security group

arjunmadan:
- now use separate parameters for creating or using an existing group
- rework test cases to indicate this

### DIFF
--- a/lib/services/awsAdapter.js
+++ b/lib/services/awsAdapter.js
@@ -42,18 +42,13 @@ function AWSAdapter (options) {
 		throw new Error(JoiValidator.error.message);
 	}
 
-	this.runInstances = function (instance, securityGroup) {
+	this.runInstances = function (instance, instanceOptions) {
 		var self   = this;
 
 		var result = _.clone(instance);
 
-		var securityGroupParams = {
-			Description : "A security group created by service-maker.",
-			GroupName   : securityGroup
-		};
-
-		return self.createSecurityGroup(securityGroupParams)
-		.then(function () {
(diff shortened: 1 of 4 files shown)


lib/services/awsAdapter.js
```diff
@@ -208,36 +203,62 @@ function AWSAdapter (options) {
 
 	};
 
-	this.createSecurityGroup = function (params) {
-		var groupName = params.GroupName;
+	this.createSecurityGroup = function (create, existing) {
```
a-type: This function name doesn't seem to describe the functionality anymore.

lib/services/awsAdapter.js
```diff
@@ -208,36 +203,62 @@ function AWSAdapter (options) {
...
+				return create;
+			})
+			.catch(function (error) {
+				return Bluebird.reject(error);
+			});
+		}
+		else {
+			var error = new Error();
+			error.name = "ValidationError";
+			error.message = "Bad request: Error with security groups.";
```
a-type: This message doesn't seem very helpful to me.

test/unit/plugins/rest_spec.js
```diff
@@ -115,7 +115,7 @@ describe("The Rest plugin", function () {
 			});
 		});
 
-		describe("with valid parameters passed - including a security group", function () {
+		describe("with valid parameters passed - creating a security group", function () {
```
a-type, line 118: How about a test for using an existing security group?

a-type:
Thanks for making those changes. Thoughts about a more descriptive error if both `create` and `existing` security groups are passed?

a-type:
That's better, thanks!

PR #103, opened 2015-08-13
Lazypolling

rrpod:
Added Lazy Polling. (Polling status is stored in cache for 10 seconds)
-Removed continuous polling every 30 seconds

### DIFF
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -8,6 +8,9 @@ var AwsAdapter      = require("../services/awsAdapter");
 var AWS             = require("aws-sdk");
 var Instance        = require("../models/Instance");
 var Bluebird        = require("bluebird");
+var Catbox          = require("catbox");
+var CatboxMongoDb   = require("catbox-mongodb");
+var CacheAdapter    = require("../services/cacheAdapter");
 
 exports.register = function (server, options, next) {
 	var ec2 = new AWS.EC2({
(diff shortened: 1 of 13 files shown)


lib/plugins/rest.js
```diff
@@ -372,7 +408,15 @@ exports.register = function (server, options, next) {
 		});
 	}
 
-	next();
+	return cacheClient.startAsync()
+	.then(function () {
+		next();
```
a-type, line 417: :+1:

lib/services/awsAdapter.js
```diff
@@ -109,15 +108,12 @@ function AWSAdapter (options) {
...
+	this.updatePollingStatus = function (result) {
 		var self = this;
 		var instanceIPAddress;
 
-		return sshAdapter.SshPolling(result.instanceID)
-		.then(function () {
-			serverLog([ "info", "AWSAdapter", "SshPolling" ], "Ssh Polling is successful");
-			return self.getPublicIPAddress(result.instanceID);
-		})
+		serverLog([ "info", "AWSAdapter", "SshPolling" ], "Ssh Polling is successful");
```
a-type: Since we're revisiting this code, I think it would be helpful to add the instance id and ip address to this log. Otherwise, it doesn't seem too useful in a live environment ("which of our concurrent instance polls was successful?")

test/unit/plugins/rest_spec.js
```diff
@@ -2918,4 +3049,71 @@ describe("The Rest plugin", function () {
 			});
 		});
 	});
+
+	describe("fails to create a Database for Cache", function () {
```
a-type: Since we're using memory now, references to databases here should probably be removed.

test/unit/services/awsAdapter_spec.js
```diff
@@ -927,349 +865,445 @@ describe("The AwsAdapter class ", function () {
...
+				expect(result.state).to.equal("failed");
+				expect(result.uri).to.equal(null);
 			});
-
 		});
 
-		describe("when waitFor times out", function () {
-			var result;
-			var awsAdapter = new AwsAdapter(awsOptions);
+		describe("faces an error while checking ssh Status", function () {
```
a-type: This describe block could be merged with the previous one,

```
facing an error while checking ssh status
   and updating the instance successfully

   and failing to update the instance

   ...
```

a-type:
Ok, thanks!

PR #105, opened 2015-08-17
Using key-pairs is now possible.

arjunmadan:
- use an existing key-pair
- use the default key-pair (it is created if it doesn't exist and the location is returned in the response header)
- create a new key-pair (the private key is stored on S3 and the URL is returned in the response header)

### DIFF
--- a/lib/plugins/rest.js
+++ b/lib/plugins/rest.js
@@ -16,17 +16,21 @@ exports.register = function (server, options, next) {
 		region : "us-east-1"
 	});
 
+	var s3 = new AWS.S3();
+
 	Bluebird.promisifyAll(ec2);
+	Bluebird.promisifyAll(s3);
+
 	Bluebird.promisifyAll(Catbox);
 
 	var instances     = options.instances || new InstanceAdapter(options.mapper);
-	var awsOptions    = { serverLog : server.log.bind(server), instances : instances, ec2 : ec2 };
+	var awsOptions    = { serverLog : server.log.bind(server), instances : instances, ec2 : ec2, s3 : s3 };
 	var catboxOptions = {
 		partition : "cachedId"
 	};
 	var cacheClient   = options.cache || new Catbox.Client(CatboxMemory, catboxOptions);
-	var awsAdapter;
 
+	var awsAdapter;
 	try {
 		awsAdapter = options.awsAdapter || new AwsAdapter(awsOptions);
 	}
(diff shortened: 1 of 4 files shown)


lib/plugins/rest.js
```diff
@@ -57,7 +63,13 @@ exports.register = function (server, options, next) {
 				return awsAdapter.runInstances(instance, instanceOptions);
 			})
 			.then(function (instance) {
-				reply(instance).created("/v1/instances/" + instance.id);
+				if (instance.keyLocation) {
+					reply(instance).created("/v1/instances/" + instance.id)
+					.header("privatekeylocation", instance.keyLocation);
```
a-type: Can this be in `PascalCase` to match the `Location` header?

a-type: Also, since this header is specific to our application, it's common practice to prefix it with `X-`.

lib/services/awsAdapter.js
```diff
@@ -261,6 +272,93 @@ function AWSAdapter (options) {
...
+
+			keyPair.name = existingKeyName;
+
+			keyPairParams = {
+				KeyNames : [ existingKeyName ]
+			};
+
+			return ec2.describeKeyPairsAsync(keyPairParams)
+			.then(function () {
+				return keyPair;
```
a-type, line 298: So an existing key pair doesn't get a location set?

lib/services/awsAdapter.js
```diff
@@ -261,6 +272,93 @@ function AWSAdapter (options) {
...
+						return Bluebird.reject(error);
+					});
+				}
+				else {
+					return Bluebird.reject(error);
+				}
+			});
+		}
+		else if (createKeyName !== undefined && existingKeyName === undefined) {
+			keyPairParams = {
```
a-type, line 322: This code, and the create code above, could be consolidated into one helper function.

lib/services/awsAdapter.js
arjunmadan, line 298: The thought behind this was that service maker wouldn't know where the private key would be. A user could use a key that only they have.

test/unit/services/awsAdapter_spec.js
```diff
@@ -148,12 +187,15 @@ describe("The AwsAdapter class ", function () {
 				.to.deep.equal(DEFAULT_SSH_RULE);
 			});
 
+			//it("creates ")
```
a-type: :octocat:

lib/services/awsAdapter.js
a-type, line 298: That does make sense, of course, now that I think about it.

a-type:
Great, thanks.

a-type:
Ok, I think that's all. Thanks.

PR #106, opened 2015-08-18
Integration tests

rrpod:


a-type:
Great. Too bad we can't run these from public PRs, but the code looks good.

